PR #1, opened 2017-09-25
Add template plugin and renaming script

goosemania:
closes #2859
https://pulp.plan.io/issues/2859

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,120 @@
+Tepmlate to create your own plugin
+==================================
+
+This is the ``plugin_template`` repository to help plugin writers
+get started and write their own plugin for `Pulp Project
+3.0+ <https://pypi.python.org/pypi/pulpcore/>`__. 
+
+Clone this repository and run the provided ``rename.py`` script to create
+a skeleton for your plugin with the name of your choice. It will contain
+``setup.py``, expected plugin layout and stubs for necessary classes and methods.
+ 
+``$ git clone https://github.com/pulp/plugin_template.git``
+
+``$ cd plugin_template``
+
+``$ ./rename.py your_plugin_name``
+ 
+Check `Plugin Writer's Guide <http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html>`__ 
+for more details and suggestions on plugin implementaion.
+
+Below are some ideas for how to document your plugin.
+
+
+All REST API examples below use `httpie <https://httpie.org/doc>`__ to
+perform the requests.
+
+Install ``pulpcore``
+--------------------
+
+Follow the `installation
+instructions <docs.pulpproject.org/en/3.0/nightly/installation/instructions.html>`__
+provided with pulpcore.
+
+Install plugin
+--------------
+
+From source
+~~~~~~~~~~~
+
+Define installation steps here.
+
+Install from PyPI
+~~~~~~~~~~~~~~~~~
+
+Define installation steps here.
+
+
+Create a repository ``foo``
+---------------------------
+
+``$ http POST http://localhost:8000/api/v3/repositories/ name=foo``
+
+Add an Importer to repository ``foo``
+-------------------------------------
+
+Add important details about your Importer and provide examples.
+
+``$ http POST http://localhost:8000/api/v3/repositories/foo/importers/plugin-template/ some=params``
+
+.. code:: json
+
+    {
+        "_href": "http://localhost:8000/api/v3/repositories/foo/importers/plugin-template/bar/",
+        ...
+    }
+
+Add a Publisher to repository ``foo``
+-------------------------------------
+
+``$ http POST http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/ name=bar``
+
+.. code:: json
+
+    {
+        "_href": "http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/bar/",
+        ...
+    }
+
+Add a Distribution to Publisher ``bar``
+---------------------------------------
+
+``$ http POST http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/bar/distributions/ some=params``
+
+Sync repository ``foo`` using Importer ``bar``
+----------------------------------------------
+
+Use ``plugin-template`` Importer:
+
+``http POST http://localhost:8000/api/v3/repositories/foo/importers/plugin-template/bar/sync/``
+
+Add content to repository ``foo``
+---------------------------------
+
+``$ http POST http://localhost:8000/api/v3/repositorycontents/ repository='http://localhost:8000/api/v3/repositories/foo/' content='http://localhost:8000/api/v3/content/plugin-template/a9578a5f-c59f-4920-9497-8d1699c112ff/'``
+
+Create a Publication using Publisher ``bar``
+--------------------------------------------
+
+Dispatch the Publish task
+
+``$ http POST http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/bar/publish/``
+
+.. code:: json
+
+    [
+        {
+            "_href": "http://localhost:8000/api/v3/tasks/fd4cbecd-6c6a-4197-9cbe-4e45b0516309/",
+            "task_id": "fd4cbecd-6c6a-4197-9cbe-4e45b0516309"
+        }
+    ]
+
+Check status of a task
+----------------------
+
+``$ http GET http://localhost:8000/api/v3/tasks/82e64412-47f8-4dd4-aa55-9de89a6c549b/``
+
+Download ``foo.tar.gz`` from Pulp
+---------------------------------
+
+``$ http GET http://localhost:8000/content/foo/foo.tar.gz``
--- a/pulp_plugin_template/app/models.py
+++ b/pulp_plugin_template/app/models.py
@@ -1,0 +1,84 @@
+"""
+Check `Plugin Writer's Guide`_ and `pulp_example`_ plugin
+implementation for more details.
+
+.. _Plugin Writer's Guide:
+    http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html
+
+.. _pulp_example:
+    https://github.com/pulp/pulp_example/
+"""
+
+from gettext import gettext as _
+from logging import getLogger
+
+from django.db import models
+
+from pulpcore.plugin.models import (Artifact, Content, ContentArtifact, RemoteArtifact, Importer,
+                                    ProgressBar, Publisher, RepositoryContent, PublishedArtifact,
+                                    PublishedMetadata)
+from pulpcore.plugin.tasking import Task
+
+
+log = getLogger(__name__)
+
+
+class PluginTemplateContent(Content):
+    """
+    The "plugin-template" content type.
+
+    Define fields you need for your new content type and
+    specify uniqueness constraint to identify unit of this type.
+
+    For example::
+
+        field1 = models.TextField()
+        field2 = models.IntegerField()
+        field3 = models.CharField()
+
+        class Meta:
+            unique_together = (field1, field2)
+    """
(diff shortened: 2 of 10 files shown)


pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # path = models.TextField(blank=False, null=False)
+    # digest = models.TextField(blank=False, null=False)
+    # objects = 
+
+    #class Meta:
+    #    unique_together = ()
+
+    @classmethod
+    def natural_key_fields(cls):
```
jortel: This seems dangerous since _unique_together_ can be either: ("name", "version")` or `(("name", "version"), ..)`

pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+class PluginTemplateContent(Content):
+    """
+    The "plugin-template" content type.
+    """
+    TYPE = 'plugin-template'
+
+    # path = models.TextField(blank=False, null=False)
+    # digest = models.TextField(blank=False, null=False)
+    # objects =
```
jortel: I think a comment like:
```
# DEFINE FIELDS HERE
```
instead of commented out fields definitions would be clearer.

pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class PluginTemplateContent(Content):
+    """
+    The "plugin-template" content type.
+    """
+    TYPE = 'plugin-template'
+
+    # path = models.TextField(blank=False, null=False)
+    # digest = models.TextField(blank=False, null=False)
```
jortel: The default is _blank=False_ and _null=False_ so just adds clutter here.

jortel:
Did you consider instead replacing all forms of _plugin-template_, we defined a set of replaceable tokens like: 

- `<NAME>` - The name.
- `<NAME-CAMEL>` - The name formatted to camel case.
- `<NAME-UNDERSCORE>` - The name formatted with underscores.
- `<NAME-DASH>` - The name formatted with dashes.

Then replaced the tokens with the formatted name.  This approach seems clearer to the plugin writer as to what needs to be replaced.  And, more straight forward for the script to do the substitution.

Examples:

```
class <NAME-CAMEL>Content(Content):
    """
    The <NAME> content type.
    """
```

```
class <NAME-CAMEL>Importer(Importer)
    """
    An importer for <NAME>Content content. 
    """

    TYPE = <NAME-UNDERSCORE>
```

```
class <NAME-CAMEL>Publisher(Publisher)
    """
    A publisher for <NAME>Content content. 
    """

    TYPE = <NAME-UNDERSCORE>
```

```
class <NAME-CAMEL>ContentViewSet(ContentViewSet):
    endpoint_name = '<NAME-DASH>'
```

goosemania:
@jortel , that's a good point.
The way it is done now the code works even without renaming. I'm not sure there is a lot of value in it though. And following your suggestion, maybe it make sense to use some template engine, like Jinja. Or it's too much for this task? :)

pulp_plugin_template/app/models.py
goosemania: Do you suggest to complete this task first? https://pulp.plan.io/issues/3025

jortel:
Hard to say about the template engine.  I haven't used any of them so not sure what would be more/less effort.

pulp_plugin_template/app/models.py
jortel: I'm not sure it needs to be done first but suggest you remove this impl here for this PR.

jortel: No, 3025 has a checklist item to vet on pulp-dev.  Suggest you just leave this in for now so you are not blocked.

setup.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+from setuptools import setup
+
+requirements = [
+    'pulpcore-plugin',
+]
+
+setup(
+    name='pulp-plugin-template',
```
asmacdo: one thing that could confuse a newbie is that this is a functioning setup.py. One benefit of switching to jinja would be that it would be clear that this file is a template for the plugin rather than the installer for the template.

rename.py
```diff
@@ -0,0 +1,105 @@
...
+    root_dir = os.path.dirname(os.path.abspath(parser.prog))
+
+    # rename python package directory
+    listed_dir = os.listdir(root_dir)
+    if TEMPLATE_SNAKE in listed_dir:
+        os.rename(os.path.join(root_dir, TEMPLATE_SNAKE),
+                  os.path.join(root_dir, pulp_plugin_name))
+
+    # replace text
+    for dir_path, dirs, files in os.walk(root_dir):
```
asmacdo: If we use a copy/rename workflow, it would be good to ignore hidden files that start with "."

rename.py
```diff
@@ -0,0 +1,105 @@
...
+TEMPLATE_DASH_SHORT = 'plugin-template'
+IGNORE_DIRS = ['.git']
+IGNORE_FILES = ['.gitignore', 'LICENSE', 'rename.py', 'flake8.cfg']
+
+
+def is_valid(name):
+    """
+    Check if specified name is compliant with requirements for it
+    """
+    return bool(re.match(r'^[a-z][0-9a-z_]{2,15}$', name))
```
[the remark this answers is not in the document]
goosemania: Will do.
There are few reasons why I suggest to put at least some limit. One limit for us is a directory name - depends on filesystem but usually max filename length is 255. We also use plugin name in class names so for readability and to fit line length :) it makes sense to ask for something relatively short I think. And 16 characters seem reasonable to me and it can be easily changed by anyone in case they need more.

pulp_plugin_template/app/viewsets.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class PluginTemplateImporterViewSet(platform.ImporterViewSet):
+    endpoint_name = 'plugin-template'
+    queryset = models.PluginTemplateImporter.objects.all()
+    serializer_class = serializers.PluginTemplateImporterSerializer
+
+
+class PluginTemplatePublisherViewSet(platform.PublisherViewSet):
+    endpoint_name = 'plugin-example'
```
asmacdo: s/plugin-example/plugin-template/

rename.py
goosemania: For now the assumption is that plugin writer doesn't do anything (except maybe just checking if template kinda works right away) before renaming, so only git-related stuff is hidden. It is ignored explicitly. Sounds ok?

pulp_plugin_template/app/viewsets.py
goosemania: good catch :)

goosemania:
Addressed most of the feedback, added docs and docstrings.

I'm on the fence on choosing between:
 - "working template even without renaming" aka class names and other objects contain `PluginTemplate` substring, simply replace by that substring
 - "explicit template string like <NAME-CAMEL>", simply replace by that substring
 - use template engine like Jinja

Currently the first option is implemented. Let me know if someone have strong opinion towards other options or have any new ideas.

README.rst
```diff
@@ -0,0 +1,115 @@
...
+        "_href": "http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/bar/",
+        ...
+    }
+
+Add a Distribution to Publisher ``bar``
+---------------------------------------
+
+``$ http POST http://localhost:8000/api/v3/repositories/foo/publishers/plugin-template/bar/distributions/ some=params``
+
+Sync repository ``foo`` using importer ``bar``
```
asmacdo: s/importer/Importer/

pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+class PluginTemplatePublisher(Publisher):
+    """
+    A Publisher for PluginTemplateContent.
+
+    A ``publish`` method should be defined.
+    It is responsible for publishing metadata and artifacts
+    which belongs to a specific repository.
+
+    For more details see Plugin Writer's Guide.
```
asmacdo: I don't think we need this in the class level docstrings. I think it makes the most sense early in the documentation, [maybe here](https://github.com/pulp/plugin_template/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R17). I also like that it is included in the module level docstring.

pulp_plugin_template/app/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+    add those fields to the Meta class keeping fields from the parent class as well.
+
+    For example::
+
+    field1 = serializers.TextField()
+    field2 = serializers.IntegerField()
+    field3 = serializers.CharField()
+
+    class Meta:
+        fields = platform.ContentSerializer.Meta.fields + ('field1', 'field2', 'field3')
```
asmacdo: Good idea to include this example, this would be easy to overlook.

rename.py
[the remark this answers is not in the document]
asmacdo: Seems reasonable. I think it is helpful to add a comment though, or else future pulp engineers might assume there is a  correctness reason that it is limited to 16.

rename.py
```diff
@@ -0,0 +1,117 @@
...
+    listed_dir = os.listdir(dst_root_dir)
+    if TEMPLATE_SNAKE in listed_dir:
+        os.rename(os.path.join(dst_root_dir, TEMPLATE_SNAKE),
+                  os.path.join(dst_root_dir, pulp_plugin_name))
+
+    # replace text
+    for dir_path, dirs, files in os.walk(dst_root_dir):
+
+        # skip directories which don't need any text replacement
+        for dir in set(IGNORE_DIRS).intersection(dirs):
```
asmacdo: I think we can get rid of this part, and the IGNORE_DIRS contant. Both __pycache__ and .git are in IGNORE_COPYTREE, so I don't think these sets will ever intersect.

goosemania: Good catch, it was needed if we did renaming in-place without copy the whole directory

README.rst
```diff
@@ -0,0 +1,118 @@
...
+Clone this repository and run the provided ``rename.py`` script to create
+a skeleton for your plugin with the name of your choice. It will contain
+``setup.py``, expected plugin layout and stubs for necessary classes and methods.
+ 
+``git clone https://github.com/pulp/plugin_template.git``
+``cd plugin_template``
+``./rename.py your_plugin_name``
+ 
+Check `Plugin Writer's Guide <http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html>`__ 
+for more details ans suggestions on plugin implementaion.
```
dkliban: s/ans/and/

README.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Below are some ideas how to document your plugin.
```
dkliban: s/ideas how/ideas for how/

pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,82 @@
...
+class PluginTemplatePublisher(Publisher):
+    """
+    A Publisher for PluginTemplateContent.
+
+    A ``publish`` method should be defined.
+    It is responsible for publishing metadata and artifacts
+    which belongs to a specific repository.
+    """
+    TYPE = 'plugin-template'
+
```
dkliban: We should mention that any fields specific to this publisher can also be specified here.

pulp_plugin_template/app/models.py
```diff
@@ -0,0 +1,82 @@
...
+class PluginTemplateImporter(Importer):
+    """
+    An Importer for PluginTemplateContent.
+
+    A ``sync`` method should be defined.
+    It is responsible for parsing metadata of the content,
+    downloading of the content and saving it to Pulp.
+    """
+    TYPE = 'plugin-template'
+
```
dkliban: Same as with the publisher, the importer for this plugin can have some additional fields.

PR #10, opened 2018-06-28
Update the template

dralley:


daviddavis:
Opened a PR to fix the pep8speaks problems that are caused by no config file: #12

PR #11, opened 2018-07-08
Fix OpenAPI schema for async responses

werwty:
closes #3668
https://pulp.plan.io/issues/3668


daviddavis:
Opened a PR to fix the pep8speaks problems that are caused by no config file: https://github.com/pulp/plugin_template/pull/12

PR #13, opened 2018-08-03
Update and add tests to the template

dralley:
* Updated the travis config/script files
* Updated the sync template for DeclarativeVersion
* Added new functionality to bootstrap.py to be able to remove anything
  between TEMPLATE_REMOVE_START and TEMPLATE_REMOVE_END tags.
* Added smash and unit test templates

closes #3773
https://pulp.plan.io/issues/3773

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 sudo: required
 # https://docs.travis-ci.com/user/trusty-ci-environment/
-dist: trusty
+dist: xenial
 language: python
 python:
     # python versions used in el7 SCL & supported fedora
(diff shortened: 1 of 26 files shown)


dralley:
We could, in theory run the travis test procedure directly on this repo.

However, I'm not sure how useful that would be.  We would need to disable the linting checks because of unused imports, and the vast majority of the smash tests would require a bit of manual intervention to add content-specific details to fix (and should be skipped).  And maybe all the smash tests should be left skipped.

In any of the above cases, it's not very useful to do so.

PR #15, opened 2018-08-30
Move plugin writer docs to template for #3772

werwty:


dralley:
@werwty This all looks really good, thank you!

My only suggestion is that in some places the docs are still using uuids.  It's not a blocker but if you could fix those before merging, that would be great.

PR #17, opened 2018-12-03
Update plugin template

dralley:
* Add mirror parameter to syncs
* Add lazy download support to declarative version
* Add a test to make sure file descriptors are not left open after a sync has completed
* Simplify content unit deletion test

### DIFF
--- a/pulp_plugin_template/tests/functional/constants.py
+++ b/pulp_plugin_template/tests/functional/constants.py
@@ -9,8 +9,9 @@ from pulp_smash.pulp3.constants import (
 )
 
 # FIXME: remove any download policies not supported by your plugin type
-DOWNLOAD_POLICIES = ['cache_only', 'immediate', 'on_demand']
-"""Allowed download policies. Defaults to immediate."""
+# If your plugin supports all download policies, you can import this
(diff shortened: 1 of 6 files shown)


pulp_plugin_template/tests/functional/constants.py
```diff
@@ -8,6 +8,10 @@
     CONTENT_PATH
 )
 
+# FIXME: remove any download policies not supported by your plugin type
+DOWNLOAD_POLICIES = ['streamed', 'immediate', 'on_demand']
```
asmacdo: since we don't support streamed or on_demand out of the box, maybe comment out this line and add a new line below:

    DOWNLOAD_POLICIES = ['immediate']

asmacdo:
Also, mind squashing? (I think you can do that with github now, but I've never tried it)

dralley:
I'll leave this for another day just in case I find anything else to add in

pulp_plugin_template/tests/functional/constants.py
rochacbruno: I would use a `tuple`

```
DOWNLOAD_POLICIES = ('immediate',)
```

PR #24, opened 2018-12-18
Replace erroneous occurance of File

mdellweg:
[noissue]


dralley:
Thank you @mdellweg !

PR #37, opened 2019-03-01
Update 'counts' response format tests

bmbouter:
Due to the breaking change in the PR below, the resonse format has
changed. No user facing features need updating, only the test assertions
change by updating the fixture data.

Required PR: https://github.com/pulp/pulpcore/pull/2

https://pulp.plan.io/issues/4283
re #4283


bmbouter:
Closing because we're going to port pulp-smash to provide the old interface and it can roll out it's breaking changes over time.

PR #39, opened 2019-04-25
Build REST API docs for publishing on rtd

asmacdo:
With `make html` and hosting the build directory with a simple server,
this PR adds a swagger docs page
http://localhost:1234/restapi.html#section/Authentication

re #4637

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -42,6 +42,7 @@ clean:
 	-rm -rf $(BUILDDIR)/*
 
 html:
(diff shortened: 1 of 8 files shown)


asmacdo:
This will not be as helpful as I originally thought, since the docs travis job includes a number of files that are not "template managed". Put another way, the `rest_api.html` file makes sense as a template managed file, but since the rest of the docs can't be clobbered, this change isn't helpful without changing the docs action of the template.

PR #40, opened 2019-05-02
Problem: auto generated REST API docs are missing

dkliban:
Solution: add everything needed to include REST API docs

closes: #4769
https://pulp.plan.io/issues/4769

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,10 +212,16 @@ TODO
 
 # Documentation
 
-Your bootstrap template comes with a set of [prepopulated docs](docs/). You can host these on readthedocs when you are ready.
+Your bootstrap template comes with a set of [prepopulated docs](docs/). You can host these on
+readthedocs when you are ready.
 
-Pulp also comes with a set of [auto API docs](meta_docs/reference/documentation.md). When your plugin is installed endpoints in the live api docs will be automatically populate.
+Pulp also comes with a set of [auto API docs](meta_docs/reference/documentation.md). When your
+plugin is installed endpoints in the live api docs will be automatically populate.
 
+When you run 'make html' command to build the docs, you must have the pulp-api running on
+localhost. The 'make html' command first downloads the OpenAPI schema for the plugin and saves it
+in ``docs/_static/api.json``. You should add this file to git. This file will then provide data
+needed to display the restapi.html page in the root of the built docs. 
 
 # Additional Topics
 
--- a/docs/_templates/restapi.html
+++ b/docs/_templates/restapi.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>REST API for Pulp 3 plugin-template Plugin</title>
+    <!-- needed for adaptive design -->
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -212,10 +212,16 @@ TODO
...
+Your bootstrap template comes with a set of [prepopulated docs](docs/). You can host these on
+readthedocs when you are ready.
 
-Pulp also comes with a set of [auto API docs](meta_docs/reference/documentation.md). When your plugin is installed endpoints in the live api docs will be automatically populate.
+Pulp also comes with a set of [auto API docs](meta_docs/reference/documentation.md). When your
+plugin is installed endpoints in the live api docs will be automatically populate.
 
+When you run 'make html' command to build the docs, you must have the pulp-api running on
+localhost. The 'make html' command first downloads the OpenAPI schema for the plugin and saves it
+in docs/_static/api.json. You should add this file to git. This file will then provide data needed
```
bmbouter: Does this need to be escaped? The _ is throwing the syntax off it seems, but I haven't seen it rendered.

docs/conf.py
```diff
@@ -11,11 +11,6 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-try:
```
bmbouter: We don't want to keep this? I actually added this recently to pulp_rpm because the left-side navigation was pretty great.

README.md
dkliban: yep. just pushed again with the text escaped.

docs/conf.py
dkliban: I am not sure I understand what this provides. This is not used by pulp-python and it's left menu seems to be the same as on pulp-rpm docs. What am I missing?

https://pulp-rpm.readthedocs.io/en/latest/index.html
https://pulp-python.readthedocs.io/en/latest/index.html

docs/_templates/restapi.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>REST API for Pulp 3 RPM Plugin</title>
```
gmbnomis: `RPM` -> `plugin-template`?

docs/conf.py
bmbouter: AIUI, when rtd builds it, it applies the theme then so the rendered version there will look the same. What won't look the same is if you try to build it locally, it will get the sphinx default theme. These themes handle navigation differently so you can't develop the docs locally in a way that will look the same in production without this.

docs/_templates/restapi.html
dkliban: thanks! i'll update it

docs/conf.py
dkliban: Makes sense. I'll bring it back.

dkliban:
@bmbouter @gmbnomis I just pushed again with the requested changes.

docs/conf.py
```diff
@@ -123,7 +123,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()] if sphinx_rtd_theme else []
+# html_theme_path = [sphinx_rtd_theme.get_html_theme_path()] if sphinx_rtd_theme else []
```
bmbouter: We want to leave this uncommented I think.

PR #42, opened 2019-05-28
Problem: no way to update Travis config for plugin

dkliban:
Solution: add a script for generating travis configs

This patch introduces a new script that plugin writers can use to generate Travis
configs for their projects.

fixes: #4784
https://pulp.plan.io/issues/4784


README.md
```diff
@@ -223,6 +223,121 @@ localhost. The 'make html' command first downloads the OpenAPI schema for the pl
...
+This repository also provides a script for generating a Travis configuration. The script should be
+run with the following command.
+
+   ``$ ./generate_travis_config.py your_plugin_name --pypi-username your_pypi_username``
+
+The default behavior enables two build stages that generate client libraries using the OpenAPI
+schema. One publishes to PyPI using ```--pypi-username``` setting and the secret environment
+variable called $PYPI_PASSWORD. The other stage publishes the client to rubygems.org and requires
+the $RUBYGEMS_API_KEY environment variable to be set. Both environment variables can be created on
+the travis-ci.com settings page for the plugin[0]. The stage that publishes tagged builds to PyPI
```
bmbouter: we need to mention the travis-ci.org option too.

README.md
```diff
@@ -223,6 +223,121 @@ localhost. The 'make html' command first downloads the OpenAPI schema for the pl
 in ``docs/_static/api.json``. You should add this file to git. This file will then provide data
 needed to display the restapi.html page in the root of the built docs. 
 
+# Travis configuration
+
+This repository also provides a script for generating a Travis configuration. The script should be
+run with the following command.
+
+   ``$ ./generate_travis_config.py your_plugin_name --pypi-username your_pypi_username``
```
bmbouter: The help text shows the `your_plugin_name` as `plugin_name` and it is the last parameter after the kwargs.

dkliban: i updated it to the same order as the help text

PR #53, opened 2019-06-17
convert template to jinja, combine interfaces

asmacdo:
Many small TODOs remain. Working with this has shown some places where the template is outdated. Rather than fix them here, I'll file issues for these where possible.

In addition, I need to rename the main command from "bootstrap" since bootstrapping is now an option.


bootstrap.py
```diff
@@ -1,112 +1,287 @@
...
 import utils
+import yaml
+
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+
+
+excluded_files = collections.defaultdict(set)
+
+
+class AppendExclude(argparse.Action):
```
asmacdo: This will be removed, it was one option for implementation, but not worth doing because we will switch to yaml config instead of CLI args

asmacdo: However, this could be useful for the update command, to skip certain files for a particular run.

asmacdo:
TIL: pep8speaks really doesn't like jinja templates that end in .py.

README.md
```diff
@@ -1,6 +1,6 @@
 <!-- TEMPLATE_REMOVE_START -->
 
-This is the ``plugin_template`` repository to help plugin writers get started and write their own
+This is the ``{{ plugin_snake_short }}`` repository to help plugin writers get started and write their own
```
dkliban: i don't think you meant to make this change

README.md
```diff
@@ -29,15 +29,15 @@ change to create a 'real' plugin.
 
 1. Clone this repository
 
-   ``$ git clone https://github.com/pulp/plugin_template.git``
+   ``$ git clone https://github.com/pulp/{{ plugin_snake_short }}.git``
```
dkliban: or this one

README.md
```diff
@@ -29,15 +29,15 @@ change to create a 'real' plugin.
...
 
-   ``$ cd plugin_template``
+   ``$ cd {{ plugin_snake_short }}``
```
dkliban: or this one

README.md
```diff
@@ -239,13 +239,13 @@ uses the same configs as the client publishing stage. The default pipeline can b
...
-$ git clone git@github.com/pulp/plugin_template
-$ cd plugin_template
+$ git clone git@github.com/pulp/{{ plugin_snake_short }}
+$ cd {{ plugin_snake_short }}
 $ # copying the requirements file is only needed if plugin was created before this file was added
 $ # to the plugin template
 $ cp doc_requirements.txt ../pulp_<plugin_name>/
 $ touch ../pulp_<plugin_name>/.travis/test_bindings.py
-$ ./generate_travis_config.py --pypi-username your_pypi_username plugin_name
+$ ./bootstrap.py --pypi-username your_pypi_username plugin_name
```
dkliban: doesn't this need --travis?

README.md
```diff
@@ -266,18 +266,18 @@ in the following order, with optional plugin provided scripts in bold:
...
-                                 [--exclude-deploy-to-pypi]
-                                 [--exclude-test-bindings]
-                                 [--exclude-deploy-client-to-pypi]
-                                 [--exclude-deploy-client-to-rubygems]
-                                 [--exclude-deploy-daily-client-to-pypi]
-                                 [--exclude-deploy-daily-client-to-rubygems]
-                                 [--pypi-username PYPI_USERNAME]
-                                 [--exclude-check-commit-message]
-                                 plugin_name
+$ ./bootstrap.py --help
```
dkliban: i don't think this helptext is accurate.

asmacdo: Would you mind if I just show the help command and remove the output?

asmacdo: Nevermind, I've just re-rendered the help text

PR #68, opened 2019-07-02
Problem: can't use a config file when updating plugin

dkliban:
Solution: add ability to use config file

re: #4928
https://pulp.plan.io/issues/4928

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -167,16 +167,36 @@ def main():
     else:
         config['plugin_snake'] = 'pulp_' + config['plugin_name']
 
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -200,6 +220,8 @@ def main():
     with open(os.path.join(plugin_root_dir, 'template_config.yml'), 'w') as outfile:
         outfile.write("# This config represents the latest values used when running the "
                       "template.\n\n")
+        config.pop('verbose')
+        config.pop('travis')
```
mdellweg: Why is travis to be excluded from the saved config?

PR #70, opened 2019-07-03
Problem: can't generate a new plugin template config

dkliban:
Solution: add command line argument for generating a config

This patch changes the name of the boostrap.py script to plugin-template. The plugin writer is now
requred to generate a template_config.yml before bootstrapping a new plugin.

Adds a --generate-config argument that generates a template_config.yml inside the plugin's root
directory. This config is then going to be used to generate Travis configuration and
scripts.

Plugin names can now be anything. The user must specify a plugin-app-label when generating a config.

This patch removes all the command line arguments that are now represented in the template config.

The settings in the template_config.yml no longer have exclude_ prefix.

The location of where the 'app' and 'tests' directories get created inside a new plugin has been
updated.

The documentation was updated to reflect the new workflow.

re: #4082
https://pulp.plan.io/issues/4082


dkliban:
@asmacdo Please checkout this PR. Feel free to merge it while I am away.

dkliban:
@mdellweg take a look at these changes.

README.md
```diff
@@ -22,26 +22,25 @@ This allows you to test your plugin at every step.
...
    and tests.
 
-   ``$ ./bootstrap.py --bootsrap your_plugin_name``
+   ``$ ./plugin-template --bootsrap plugin_name plugin_app_label``
 
-   **NOTE** : Whatever you choose for `your_plugin_name` will be prefixed with `pulp_`.
-   Therefore, for this argument it is best to just provide the content type
-   which you would like to support, e.g. `rubygem` or `maven`.
+   **NOTE** : The `plugin_app_label` should identify the content type which you would like to support, e.g. `rubygem` or `maven`.
+   The `plugin_name` is usually `pulp_` or `pulp-` prepended to the `plugin_app_label`, e.g. `pulp_maven`.
```
mdellweg: Do we really want to give away that freedom? Also i do not see the benefit. I think we cannot have `pulp_perl` _and_ `pulp-perl` at the same time. Also, we do not want that, i guess.

utils.py
```diff
@@ -33,3 +33,10 @@ def to_dash(name):
     Convert plugin name from snake case to dash representation
     """
     return name.replace('_', '-')
+
+
+def to_snake(name):
+    """
+    Convert plugin name from snake case to dash representation
```
mdellweg: It's the other way around.

README.md
```diff
@@ -298,19 +297,21 @@ optional arguments:
 
   --verbose             Include more output.
 
-  --pypi-username PYPI_USERNAME
+
+The following config values are available:
+  pypi-username
```
mdellweg: Why should the dashes be missing here. If, because they can also specified in a config elsewhere, where is that?

README.md
dkliban: We already have a plugin that's called pulp-certguard[0]. I don't expect anyone to try to create pulp_certguard. I don't think the plugin template needs to guard against this. 

https://github.com/pulp/pulp-certguard/

utils.py
dkliban: Thank you ... I'll update.

README.md
dkliban: The dashes are removed because these parameters are no longer passed in on the command line. They only appear in the template_config.yml file inside the plugin repository.

mdellweg: Still, imho this README is unclear about where to specify that stuff.

asmacdo: Adding generate-config to the walkthrough will help

plugin-template
```diff
@@ -0,0 +1,265 @@
...
+    'plugin_dash_short': None,
+    'plugin_name': None,
+    'plugin_snake': None,
+    'plugin_snake_short': None,
+    'pypi_username': None
+}
+
+def main():
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                     description='Generate a .travis.yml and .travis directory'
```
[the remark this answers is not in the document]
dkliban: I'll update it with a better message.

plugin-template
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
dkliban: Just didn't think the .py was needed.

plugin-template
```diff
@@ -0,0 +1,265 @@
...
+
+    try:
+        with open(plugin_config_path) as config_file:
+            try:
+                config_in_file = yaml.safe_load(config_file)
+                if config_in_file:
+                    # TODO: validate config
+                    config = config_in_file
+                    # Add any missing value from the list of defaults
+                    for key, value in DEFAULT_SETTINGS.items():
```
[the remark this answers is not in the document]
dkliban: Yes it is needed because there may have been new settings added since the plugin author last used plugin-template.

plugin-template
```diff
@@ -0,0 +1,269 @@
...
+                    print("\nLoaded plugin template config from "
+                          "{path}/template_config.yml.\n".format(path=plugin_root_dir))
+                    if args.verbose:
+                        print("\nUsing the following config:\n")
+                        pp = pprint.PrettyPrinter(indent=4)
+                        pp.pprint(config)
+                        print("\n")
+            except yaml.YAMLError as exc:
+                print(exc)
+                exit()
```
gmbnomis: This should return with an error, i.e. we should do `return 2` or so like below

.travis.yml
```diff
@@ -3,4 +3,4 @@ python:
   - "3.7"
 install: pip install -r test_requirements.txt
 script:
-  ./bootstrap.py --bootstrap --all catdog && cd ../pulp_catdog && black --diff --check --exclude docs/conf.py .
+  ./plugin-template --bootstrap --all pulp_catdog catdog && cd ../pulp_catdog && black --diff --check --exclude docs/conf.py .
```
gmbnomis: Isn't `bootstrap` just one of the sections included in `--all`? It looks redundant(?)

plugin-template
```diff
@@ -0,0 +1,269 @@
...
+    Args:
+        plugin_name (str): the full name of the plugin repository name (e.g. pulp_file)
+        plugin_app_label (str): the Django app label for the plugin - usually the part after
+            pulp_ or pulp-
+
+    Returns:
+        config (dict): a dictionary containing a default template configuration for the plugin
+    """
+    config = {}
+    config['plugin_name'] = plugin_app_label
```
gmbnomis: This looks odd. I would have expected:
```python
config['plugin_name'] = plugin_name
config['plugin_app_label'] = plugin_app_label
```

README.md
```diff
@@ -52,6 +51,100 @@ as they will make continuous verification of your plugin's functionality much ea
...
 plugin-writer intervention is required.
 
+# Generate a template_config.yaml for an existing plugin
+
+The ``plugin-template`` script can update a plugin with the latest Travis CI configuration. Plugin
+authors can select which parts of the CI pipeline are included by providing a template_config.yaml
+in the root of the plugin repository. Run the following command to generate a template_config.yaml
+with all the defaults.
+
+    ``$ ./plugin-template --generate-config plugin_name plugin_app_label``
```
gmbnomis: I think it is not completely clear from the description that this command is not used in the root of the plugin but in the root of plugin_template and expects the plugin at `../<plugin_name>` (i.e. it generates `../<plugin_name>/template_config.yml`.

plugin-template
```diff
@@ -0,0 +1,269 @@
...
+                    if args.verbose:
+                        print("\nUsing the following config:\n")
+                        pp = pprint.PrettyPrinter(indent=4)
+                        pp.pprint(config)
+                        print("\n")
+            except yaml.YAMLError as exc:
+                print(exc)
+                exit()
+    except FileNotFoundError:
+        print("\nCould not find a plugin template config at {path}/template_config.yml.\n".format(
```
gmbnomis: Proposal: Add `Using default configuration` or so to make clear that this is not an error.

README.md
```diff
@@ -52,6 +51,100 @@ as they will make continuous verification of your plugin's functionality much ea
...
+
+This is going to generate a template_config.yaml with the following settings:
```
gmbnomis: If I get the code correctly, `--generate-config` will not generate a new default config when  a config file is already existing (it will parse the existing config and add defaults for new settings that are not part of the file).

I like this behavior, but I think it should be made clearer here. Judging from the docs, I would have expected it to just generate a new config file.

plugin-template
```diff
@@ -0,0 +1,269 @@
...
+                not config['exclude_deploy_to_pypi']:
+            print("If PyPI scripts are included, PyPI username is required.")
+            return 2
+
+        if not config['exclude_test_bindings']:
+            if not os.path.isfile(os.path.join(plugin_root_dir, '.travis', 'test_bindings.py')):
+                print("If bindings are not excluded, `test_bindings.py` must exist.")
+                return 2
+
+    if args.generate_config:
```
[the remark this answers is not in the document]
dkliban: --all will generate it on line 145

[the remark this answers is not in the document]
gmbnomis: I see, I should have had a more detailed look at what the code actually does (I thought `--generate-config` behaves just like the section options). 

But it does not and the current behavior is:

**plugin directory does not exist:**

`./plugin-template --generate-config pulp-test test` -> Exception (can't write config as directory does not exist)

`./plugin-template pulp-test test` -> Exception (can't write config as directory does not exist)

`./plugin-template --generate-config --bootstrap pulp-test test` -> Exception (can't write config as directory does not exist)

**plugin directory does exist:**

`./plugin-template --generate-config pulp-test test` -> Generates config

`./plugin-template pulp-test test` -> Generates config!

`./plugin-template --generate-config --bootstrap pulp-test test` -> Generates config and silently ignores `--bootstrap`

IMHO, this does not follow the principle of least surprise. As said, I thought that `--generate-config` just behaves like the other sections (and creates the directory automatically if needed)

plugin-template
dkliban: This is due to the names that were used when the plugin template was first created. I'll try to change this to be more straight forward.

README.md
dkliban: I'll add this clarification.

.travis.yml
dkliban: Good point. I'll update this.

plugin-template
dkliban: I'll add that language.

dkliban:
@asmacdo @gmbnomis @mdellweg This PR is ready for re-review. I updated the description with all the changes.

plugin-template
asmacdo: Optional: Could the exception mention  --generate config?

plugin-template
```diff
@@ -0,0 +1,263 @@
...
+    """
+    section_template_dir = "templates/{name}".format(name=name)
+    env = Environment(
+        loader=FileSystemLoader([section_template_dir, 'templates'])
+    )
+
+    files_templated = 0
+    files_copied = 0
+    for relative_path in generate_relative_path_set(section_template_dir):
+        destination_relative_path = relative_path.replace('plugin_name', config['plugin_snake'])
```
asmacdo: This is nice. :+1:

dkliban:
@gmbnomis @mdellweg: I am going to merge this PR, however, if you find any problems or have suggestions for improvements, please open up an issue on pulp.plan.io

PR #77, opened 2019-07-09
Improve before_install.sh

gmbnomis:
- Use bash and the more secure `set -mveuo pipefail` (as in other scripts)

- Use shallow clones and shallow fetch to speed up the CI run (Especially
  pulpcore takes a long time to clone completely.  We don't need any history
  when building in the CI.)

- get rid of pushd/popd (bashism that is not needed here)

https://pulp.plan.io/issues/5078
fixes #5078


gmbnomis:
I updated the commit to close https://pulp.plan.io/issues/5078

PR #86, opened 2019-07-18
Creating functional test script hook

daviddavis:


templates/travis/.travis/script.sh.j2
```diff
@@ -90,11 +91,11 @@ coverage run $(which django-admin) runserver 24817 --noreload >> ~/django_runser
 wait_for_pulp 20
 
 # Run functional tests
-pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional || show_logs_and_return_non_zero
 {% if plugin_name != 'pulpcore' %}pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional || show_logs_and_return_non_zero
```
gmbnomis: Idea: Put this `pytest` call into the `else` clause of `if [ -x $FUNC_TEST_SCRIPT ]; then` below.

This would allow plugins to override how to run the plugin's functional test (and still provide a default for the common case). Additionally, we can get rid of magic plugin name `if`s.

PR #93, opened 2019-08-07
Remove maria/mysql from installer

CodeHeeler:
Remove the pulp_db_type variable from installer and template

fixes #5130
https://pulp.plan.io/issues/5130

### DIFF
--- a/templates/travis/.travis/playbook.yml.j2
+++ b/templates/travis/.travis/playbook.yml.j2
@@ -4,7 +4,7 @@
 - hosts: all
   pre_tasks:
     - name: Load DB variables
-      include_vars: '{% raw %}{{ pulp_db_type }}{% endraw %}.yml'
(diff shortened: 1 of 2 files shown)


templates/travis/.travis/playbook.yml.j2
```diff
@@ -23,7 +23,6 @@
       secret_key: 'secret'
       databases:
         default:
-          ENGINE: {% raw %}"{{ pulp_db_backend }}"{% endraw %}
```
dkliban: This line needs to set the ENGINE to a string that represents postgresql.

dkliban: Please disregard the previous comment. The value is already set as a default here: https://github.com/pulp/ansible-pulp/blob/master/roles/pulp-database/defaults/main.yml#L9

PR #100, opened 2019-09-03
Fix test stage

mikedep333:
Problem: Travis does no builds when the build matrix has 1 build
    
    build matrix for us == Python versions * (docs + bindings + pulp)
    
    Solution: Add a stub implicit job called "test" when there is only
    the TEST=pulp build.
    
    [noissue]
    
    Fixes a regression introduced as part of:
    re: #5004
    https://pulp.plan.io/issues/5004
    "migrate travis plugin-template to installing pulp in containers"

PR also includes a revert for the previous attempted to fix: Which worked for pulp_deb (1 build) but broke other plugins (2+ builds).

### DIFF
--- a/templates/travis/.travis.yml.j2
+++ b/templates/travis/.travis.yml.j2
@@ -57,6 +57,16 @@ after_failure:
   - sudo kubectl logs -l app=pulp-worker --tail=10000
 jobs:
   include:
+{%- if not test_bindings and not docs_test %}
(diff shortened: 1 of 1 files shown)


mikedep333:
PRs generated with this:

https://github.com/pulp/pulp_deb/pull/107
https://github.com/pulp/pulp_file/pull/270 (docs build failing due to an issue that seems unrelated to this)

PR #103, opened 2019-09-09
Problem: Containers are built when publishing to PyPI, etc.

mikedep333:
Solution: Disable install phase for all deploy & publish steps.

fixes: #5409
Containers are built when testing docs and publishing to PyPI
https://pulp.plan.io/issues/5409

Note: I am only able to partially test this. On Travis mikeded333/pulpcore, the publishing task does not attempt to run the install script. However, it cannot push the docs due to lack of credentials (in Travis env vars.)
https://travis-ci.com/mikedep333/pulpcore/jobs/231988570#L565

### DIFF
--- a/templates/travis/.travis.yml.j2
+++ b/templates/travis/.travis.yml.j2
@@ -69,6 +69,7 @@ jobs:
 {%- endif %}
     {%- if deploy_to_pypi %}
     - stage: deploy-plugin-to-pypi
+      install: skip
       script: bash .travis/publish_plugin_pypi.sh
       if: tag IS present
     {%- endif %}


templates/travis/.travis.yml.j2
```diff
@@ -69,17 +69,20 @@ jobs:
 {%- endif %}
     {%- if deploy_to_pypi %}
     - stage: deploy-plugin-to-pypi
+      install: skip
       script: bash .travis/publish_plugin_pypi.sh
       if: tag IS present
     {%- endif %}
     {%- if plugin_name == 'pulpcore' %}
     - stage: publish-beta-docs
+      install: skip
```
dkliban: we need the rest api for building the docs (rest api docs)

templates/travis/.travis.yml.j2
```diff
@@ -104,6 +109,7 @@ jobs:
     {%- endif %}
     {%- if deploy_client_to_rubygems %}
     - stage: publish-client-gem
+      install: skip
```
dkliban: we need the rest api here also

templates/travis/.travis.yml.j2
```diff
@@ -96,6 +100,7 @@ jobs:
     {%- endif %}
     {%- if deploy_daily_client_to_pypi %}
     - stage: publish-daily-client-pypi
+      install: skip
```
dkliban: we need the rest api here also

templates/travis/.travis.yml.j2
```diff
@@ -88,6 +91,7 @@ jobs:
     {%- endif %}
     {%- if deploy_daily_client_to_rubygems %}
     - stage: publish-daily-client-gem
+      install: skip
```
dkliban: we need the rest api here also

templates/travis/.travis.yml.j2
```diff
@@ -69,17 +69,20 @@ jobs:
...
     {%- if plugin_name == 'pulpcore' %}
     - stage: publish-beta-docs
+      install: skip
       script: bash .travis/publish_docs.sh rc
       env:
         - DB=postgres
         - TEST=docs
       if: tag IS present
     - stage: publish-continuous-docs
+      install: skip
```
dkliban: we need the rest api here also

templates/travis/.travis.yml.j2
```diff
@@ -112,6 +118,7 @@ jobs:
     {%- endif %}
     {%- if deploy_client_to_pypi %}
     - stage: publish-client-pypi
+      install: skip
```
dkliban: This is good!

dkliban:
pulpcore is the only project that needs the rest api to build the docs. the other projects use a static asset in the repo to build the docs. the docs publishing jobs only run for pulpcore already. so the only thing that needs changing is that we don't need to spin up the rest api to publish the package to PyPI.

PR #111, opened 2019-09-24
Problem: plugins need to install addtl test deps in the pulp-api cont…

mikedep333:
…ainer

solution: Copy the env vars from script.sh to before_script.sh,
and move the common install commands to script.sh

fixes: #5497


gmbnomis:
@mikedep333 I tested this with a custom `.travis/post_before_script.sh` (https://github.com/gmbnomis/pulp_cookbook/tree/test_pulp_template_pr_111). Works like a charm!

PR #112, opened 2019-09-28
Travis: Build on tag uses PyPI modules

gmbnomis:
When building a release on a tag, the Travis build should ensure that the
plugin works with released versions of the pulpcore and pulpcore-plugin
modules. This should catch issues like https://pulp.plan.io/issues/5509.

If the Travis build is for a tag, do not use the master versions of
these modules, but install from PyPI instead.

ref 4386
https://pulp.plan.io/issues/4386

### DIFF
--- a/templates/travis/.travis/before_install.sh.j2
+++ b/templates/travis/.travis/before_install.sh.j2
@@ -68,27 +68,32 @@ if [ -n "$PULP_PR_NUMBER" ]; then
 fi
 {% endif %}
 
-git clone --depth=1 https://github.com/pulp/pulpcore-plugin.git
+# When building a (release) tag, we don't need the development modules for the
+# build (they will be installed as dependencies of the plugin).
+if [ -z "$TRAVIS_TAG" ]; then
 
-if [ -n "$PULP_PLUGIN_PR_NUMBER" ]; then
-  cd pulpcore-plugin
-  git fetch --depth=1 origin +refs/pull/$PULP_PLUGIN_PR_NUMBER/merge
-  git checkout FETCH_HEAD
-  cd ..
+  git clone --depth=1 https://github.com/pulp/pulpcore-plugin.git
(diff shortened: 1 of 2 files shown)


templates/travis/.travis/before_install.sh.j2
```diff
@@ -68,27 +68,32 @@ if [ -n "$PULP_PR_NUMBER" ]; then
...
 
-git clone --depth=1 https://github.com/PulpQE/pulp-smash.git
 
-if [ -n "$PULP_SMASH_PR_NUMBER" ]; then
-  cd pulp-smash
-  git fetch --depth=1 origin +refs/pull/$PULP_SMASH_PR_NUMBER/merge
-  git checkout FETCH_HEAD
-  cd ..
-fi
+  git clone --depth=1 https://github.com/PulpQE/pulp-smash.git
```
mikedep333: I agree with this design, but
1. Right now, the CI will fail with this I believe. Pulp-Smash's master branch has the container (kubectl) support. Pulp-Smash needs to do a release.
2. In general, we're going to need to do a pulp-smash release at the time we release pulpcore & pulpcore-plugin.

templates/travis/.travis/install.sh.j2
```diff
@@ -55,7 +58,20 @@ else
...
+if [ -n "$TRAVIS_TAG" ]; then
+  # Install the plugin only and use published PyPI packages for the rest
+  cat > vars/vars.yaml << VARSYAML
+---
+images:
+  - ${PLUGIN}-${TAG}:
+      image_name: $PLUGIN
+      tag: $TAG
+      plugins:
+        - pulp-certguard
```
mikedep333: I agree with this design, but
1. In general, we need to do a pulp-certguard release at the time we release pulpcore & pulpcore-plugin.

PR #114, opened 2019-10-01
include additional plugins on travis

fabricio-aguiar:
closes #5500
https://pulp.plan.io/issues/5500

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -15,6 +15,7 @@ from jinja2 import Environment, FileSystemLoader
 
 
 DEFAULT_SETTINGS = {
+    'additional_plugins': [],
     'black': True,
     'check_commit_message': True,
     'coverage': False,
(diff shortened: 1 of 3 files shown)


templates/travis/.travis/install.sh.j2
```diff
@@ -69,6 +76,8 @@ images:
       plugins:
         - pulp-certguard
```
dkliban: Let's remove pulp-certguard and have it be provided by the additional_plugins config for each plugin.

plugin-template
```diff
@@ -15,6 +15,7 @@ from jinja2 import Environment, FileSystemLoader
 
 
 DEFAULT_SETTINGS = {
+    'additional_plugins': [],
```
fabricio-aguiar: @dkliban what do you think about making `pulp-certguard` default?
`'additional_plugins': ['pulp-certguard'],`

dkliban: Let's leave it empty for now. However, when we update the template_config.yml for pulpcore, we definitely want to add 'pulp-certguard' to hte list there.

PR #116, opened 2019-10-03
Fix versioning for daily client bindings

mdellweg:
fixes #5536
https://pulp.plan.io/issues/5536

### DIFF
--- a/templates/travis/.travis/publish_client_gem.sh.j2
+++ b/templates/travis/.travis/publish_client_gem.sh.j2
@@ -17,6 +17,8 @@ export DESCRIPTION="$(git describe --all --exact-match `git rev-parse HEAD`)"
 if [[ $DESCRIPTION == 'tags/'$REPORTED_VERSION ]]; then
   export VERSION=${REPORTED_VERSION}
 else
+  # Daily publishing of development version (ends in ".dev" reported as ".dev0")
+  [ "${REPORTED_VERSION%.dev*}" != "${REPORTED_VERSION}" ] || exit 1
   export EPOCH="$(date +%s)"
   export VERSION=${REPORTED_VERSION}${EPOCH}
 fi
@@ -25,7 +27,7 @@ export response=$(curl --write-out %{http_code} --silent --output /dev/null http
 
 if [ "$response" == "200" ];
 then
-    exit
+  exit
 fi
 
 cd
--- a/templates/travis/.travis/publish_client_pypi.sh.j2
+++ b/templates/travis/.travis/publish_client_pypi.sh.j2
@@ -15,15 +15,17 @@ export DESCRIPTION="$(git describe --all --exact-match `git rev-parse HEAD`)"
 if [[ $DESCRIPTION == 'tags/'$REPORTED_VERSION ]]; then
   export VERSION=${REPORTED_VERSION}
 else
+  # Daily publishing of development version (ends in ".dev" reported as ".dev0")
+  [ "${REPORTED_VERSION%.dev*}" != "${REPORTED_VERSION}" ] || exit 1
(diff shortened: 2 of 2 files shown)


templates/travis/.travis/publish_client_gem.sh.j2
```diff
@@ -17,15 +17,17 @@ export DESCRIPTION="$(git describe --all --exact-match `git rev-parse HEAD`)"
 if [[ $DESCRIPTION == 'tags/'$REPORTED_VERSION ]]; then
   export VERSION=${REPORTED_VERSION}
 else
+  # Daily publishing of development version (ends in ".dev")
+  [ "${REPORTED_VERSION%.dev}" != "${REPORTED_VERSION}" ] || exit 1
   export EPOCH="$(date +%s)"
-  export VERSION=${REPORTED_VERSION}${EPOCH}
+  export VERSION=${REPORTED_VERSION}.${EPOCH}
```
mikedep333: The test condition exits the entire script if the string does not end in .dev, correct?

So when we have a REPORTED_VERSION like:
3.0.0rc6.dev

Instead of it being listed as:
3.0.0rc6.dev1568383024
It will be listed as:
3.0.0rc6.dev.1568383024

Is that correct?

Because the example in [#5536](https://pulp.plan.io/issues/5536) does not list ".dev".

templates/travis/.travis/publish_client_gem.sh.j2
```diff
@@ -17,15 +17,17 @@ export DESCRIPTION="$(git describe --all --exact-match `git rev-parse HEAD`)"
 if [[ $DESCRIPTION == 'tags/'$REPORTED_VERSION ]]; then
   export VERSION=${REPORTED_VERSION}
 else
+  # Daily publishing of development version (ends in ".dev")
+  [ "${REPORTED_VERSION%.dev}" != "${REPORTED_VERSION}" ] || exit 1
```
mikedep333: The test condition exits the entire script if the string does not end in .dev, correct?

And the test condition should never be reached for tagged releases versions, because of the original "if" condition?

mdellweg: That is the plan, but i have it in place for debian, and it does not work as expected. I can just remove this check, and we only add the additional dot. That would at the very least give us versions that grow in the right order.
The original problem is, that when this script runs daily and the version is reported as `1.2.3rc5` it will published as `1.2.3.rc5201911111111` which will be bigger than `1.2.3.rc6`. The dot alone should help there. Also any non digit should to.

templates/travis/.travis/publish_client_gem.sh.j2
mdellweg: Yes. The problem occurs, when it is _not_ ending in `.dev`.

mdellweg: It turns out that the plugin version is reported as `"version": "2.0.0b3.dev0"` while it is specified `__version__ = "2.0.0b3.dev"` in the plugin.
This is in the source box; in the sandbox install this does not happen.

mikedep333: @mdellweg So the source box installs nightly versions, and therefore the 0 gets appended to the letter at the end.

But the sandbox installs stable releases, which already has numbers at the end, so it appends nothing?

mdellweg:
ping @dkliban

templates/travis/.travis/publish_client_gem.sh.j2
mdellweg: I had to remove the dot again, as python really did not like it.

PR #123, opened 2019-10-08
Add 'pulp_settings' template config

dkliban:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -37,6 +37,7 @@ DEFAULT_SETTINGS = {
     'plugin_name': None,
     'plugin_snake': None,
     'plugin_app_label': None,
+    'pulp_settings': None,
     'pypi_username': None,
     'travis_notifications': None,
 }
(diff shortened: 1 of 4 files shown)


templates/travis/.travis/install.sh.j2
```diff
@@ -109,7 +109,11 @@ spec:
     username: pulp
     password: pulp
     admin_password: pulp
-  content_host: $(hostname):24816
```
mikedep333: So we will not set this for all plugins anymore? It is a temporary measure, but still.

plugin-template
```diff
@@ -37,6 +37,7 @@ DEFAULT_SETTINGS = {
     'plugin_name': None,
     'plugin_snake': None,
     'plugin_app_label': None,
+    'pulp_settings': None,
```
bmbouter: I thought this would get the content_host var?

dkliban: It's up to the plugin author to provide that setting if the plugin needs it.

PR #132, opened 2019-10-28
Add missing ; to an if statement

dkliban:
[noissue]


daviddavis:
`;` is not required if `then` is on the next line.

dkliban:
Very true @daviddavis. Please help me understand why this was happening? https://travis-ci.org/pulp/pulp_ansible/jobs/603471681#L259

PR #134, opened 2019-10-28
Set content_origin automatically

bmbouter:
The plugin_template sets `content_origin` as http://$(hostname):24816`.
This aligns with the current way services are tested on Travis

https://pulp.plan.io/issues/5629
re #5629

### DIFF
--- a/templates/travis/.travis/script.sh.j2
+++ b/templates/travis/.travis/script.sh.j2
@@ -15,6 +15,10 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 
 if [ "$TEST" = 'docs' ]; then
+  {% if plugin_name == 'pulpcore' %}
+    export PULP_CONTENT_ORIGIN=http://$(hostname):24816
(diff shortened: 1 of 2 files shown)


templates/travis/.travis/script.sh.j2
```diff
@@ -13,6 +13,7 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 # Gets set in .travis/settings.yml, but doesn't seem to inherited by
 # this script.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+export PULP_CONTENT_ORIGIN=http://$(hostname):24816
```
bmbouter: This worked for me when I applied it to pulpcore, but I'm concerned about 2 things:

1) should this be guarded by the if == 'docs' below so it only applies to those builds which don't "inherit the settings"
2) should this be guarded by an if == 'pulpcore' to only set it on pulpcore?

dkliban: Yes to both.

daviddavis:
Closing in favor of https://github.com/pulp/plugin_template/pull/135

Basically, we don't need to set content_origin since it's automatically set by the pulp_operator.

PR #137, opened 2019-11-09
Set CONTENT_ORIGIN before calling django-admin runserver

daviddavis:
ref #5686


daviddavis:
Looks like this change is not needed. Adding WIP until I figure out why the docs publish job needs CONTENT_ORIGIN to manually be set but bindings don't.

PR #138, opened 2019-11-09
Only build continuous docs for master

daviddavis:
Dependabot opens PRs against non-master branches which triggers the continuous docs publish stage. We probably should only be publishing docs for master.


bmbouter:
This looks good to me. Thanks @daviddavis

PR #140, opened 2019-11-12
Fix typo

mdellweg:
[noissue]

Was there a wrong keyboard layout?


templates/bootstrap/plugin_name/app/serializers.py.j2
```diff
@@ -72,7 +72,7 @@ class {{ plugin_camel_short }}RemoteSerializer(platform.RemoteSerializer):
         model = models.{{ plugin_camel_short }}Remote
 
 
-class {{ plugin_camel_short }}RepoairoeySerializer(platform.RepoairoeySerializer):
```
dralley: Probably my fingers got misplaced on the keys

daviddavis: Hopefully you can get some rest tomorrow after we branch 3.0 :)

PR #142, opened 2019-11-15
Fix unbound variable error

daviddavis:
Repos that don't have PERFORMANCE_TEST set, they get:

```
.travis/script.sh: line 112: PERFORMANCE_TEST: unbound variable
```


mikedep333:
Good thing we have `set -u` up top :) It helps catch bugs.

PR #162, opened 2020-01-03
Run tests for newly generated plugin

fabricio-aguiar:
https://pulp.plan.io/issues/5925
closes #5925


.travis/pre_before_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+pip install -r test_requirements.txt
+./plugin-template --generate-config --plugin-app-label catdog pulp_catdog
+mkdir ../pulp_catdog/.travis
+touch ../pulp_catdog/.travis/test_bindings.py
+echo 'pypi_username: the_pypi_user' >> ../pulp_catdog/template_config.yml
+./plugin-template --all pulp_catdog
+mv ../pulp_catdog/*.txt .
+mv ../pulp_catdog/.travis/before_install.sh ../pulp_catdog/.travis/old_before_install.sh
```
fabricio-aguiar: I had to make some changes on `before_install.sh`, so I'm not using the generated `before_install.sh`.
This is the change that most worries me, suggestions are very welcome!

.travis/before_install.sh
```diff
@@ -0,0 +1,116 @@
...
+
+if [ -n "$PULP_PR_NUMBER" ]; then
+  cd pulpcore
+  git fetch --depth=1 origin +refs/pull/$PULP_PR_NUMBER/merge
+  git checkout FETCH_HEAD
+  cd ..
+fi
+
+rm ./pulpcore/containers/images/pulp/container-assets/pulp-api
+mv ./plugin_template/.travis/pulp-api ./pulpcore/containers/images/pulp/container-assets/pulp-api
```
mikedep333: So why are we replacing this?

.travis.yml
```diff
@@ -1,11 +1,44 @@
+# WARNING: DO NOT EDIT!
+#
+# This file was generated by plugin_template, and is managed by bootstrap.py. Please use
+# bootstrap.py to update this file.
```
daviddavis: Is this file maintained by bootstrap.py? If not, we should remove this header.

.travis/before_install.sh
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
+
+# WARNING: DO NOT EDIT!
+#
+# This file was generated by plugin_template, and is managed by bootstrap.py. Please use
+# bootstrap.py to update this file.
```
daviddavis: Same here.

.travis/before_install.sh
fabricio-aguiar: for running `django-admin makemigrations catdog`
https://github.com/pulp/plugin_template/blob/master/.travis/pulp-api#L14

.travis.yml
fabricio-aguiar: it isn't, I just copied from pulp_file, but forgot to remove it

PR #167, opened 2020-01-14
Add option to use pulp-fixtures docker container

daviddavis:
fixes #5913
https://pulp.plan.io/issues/5913

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -25,6 +25,7 @@ DEFAULT_SETTINGS = {
     'deploy_daily_client_to_pypi': True,
     'deploy_daily_client_to_rubygems': True,
     'deploy_to_pypi': True,
+    'docker_fixtures': False,
     'docs_test': True,
     'pydocstyle': True,
     'test_bindings': True,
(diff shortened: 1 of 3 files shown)


plugin-template
```diff
@@ -25,6 +25,7 @@ DEFAULT_SETTINGS = {
     'deploy_daily_client_to_pypi': True,
     'deploy_daily_client_to_rubygems': True,
     'deploy_to_pypi': True,
+    'docker_fixtures': False,
```
fabricio-aguiar: Nice!

templates/travis/.travis/before_install.sh.j2
```diff
@@ -51,6 +51,15 @@ black --check --diff .
 # Lint code.
 flake8 --config flake8.cfg
 
+{%- if docker_fixtures -%}
+# run fixtures container and update pulp-smash config
+docker run -d -p 8000:80 quay.io/pulp/pulp-fixtures:latest
```
fabricio-aguiar: :heart:

mikedep333: Yeah, I agree with this approach for launching the container. It's the least work vs putting it in k8s.

PR #172, opened 2020-01-17
Run bindings test

fabricio-aguiar:
Run bindings test on plugin_template
https://travis-ci.org/pulp/plugin_template/jobs/638451302


fabricio-aguiar:
we are trying to just move to use the bindings in our functional tests

PR #182, opened 2020-02-11
Introducing github actions

fabricio-aguiar:
https://pulp.plan.io/issues/6067
closes #6067

GH Actions:
https://github.com/fabricio-aguiar/plugin_template/runs/438586416?check_suite_focus=true

### DIFF
--- a/.github/before_install.sh
+++ b/.github/before_install.sh
@@ -2,8 +2,8 @@
 
 set -mveuo pipefail
 
-export PRE_BEFORE_INSTALL=$TRAVIS_BUILD_DIR/.travis/pre_before_install.sh
-export POST_BEFORE_INSTALL=$TRAVIS_BUILD_DIR/.travis/post_before_install.sh
+export PRE_BEFORE_INSTALL=$GITHUB_WORKSPACE/.github/pre_before_install.sh
+export POST_BEFORE_INSTALL=$GITHUB_WORKSPACE/.github/post_before_install.sh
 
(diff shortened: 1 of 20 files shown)


dralley:
Looks good, although we won't want to keep Travis running for long afterwards due to the duplication of the scripts.

.travis/before_install.sh
```diff
@@ -1,12 +1,5 @@
 #!/usr/bin/env bash
 
-# WARNING: DO NOT EDIT!
-#
-# This file was generated by plugin_template, and is managed by bootstrap.py. Please use
-# bootstrap.py to update this file.
-#
-# For more info visit https://github.com/pulp/plugin_template
-
```
mdellweg, line 9: Why should this header be removed?
The reference to `bootstrap.py` is wrong though.

mdellweg:
The scripts copied from travis to gh-actions look mostly identical to me.
Would it be feasible to make them ci-agnostic and put them in a common location?

.travis/before_install.sh
fabricio-aguiar: this is not from the template, it is the CI config for `plugin_template` itself, and it has some differences from the one generated from template.

fabricio-aguiar:
> The scripts copied from travis to gh-actions look mostly identical to me.
> Would it be feasible to make them ci-agnostic and put them in a common location?

This PR started as PoC, and many things changed after I started it, I believe we will remove travis entirely

fabricio-aguiar:
Waiting github actions features

PR #202, opened 2020-03-26
Problem: Trailing underscores in branch names breaks CI

mikedep333:
Solution: Use ${TAG} syntax and quote them. So that they
are cast as strings, not numbers.

fixes: #6405
https://pulp.plan.io/issues/6405
Trailing underscores in branch names breaks CI


mikedep333:
Test: https://github.com/pulp/pulpcore/pull/614

PR #203, opened 2020-03-31
Problem: Plugins cannot modify the pulp-content containers easily

mikedep333:
because there are 2 of them.

Solution: Use only 1 pulp-content pod.

We can make their # configurable if there is a need in the future.

[noissue]

re: #4664
As a user, I can use a RHSMCertGuard
https://pulp.plan.io/issues/4664


bmbouter:
This looks good to me! I'm going to try it out on my pulp-certguard branch by applying the diff manually, and I'll post back what I find.

mikedep333:
The test PR:
https://github.com/pulp/pulp-certguard/pull/48

PR #204, opened 2020-03-31
Adds pulp_certguard-client bindings for pulpcore CI

dkliban:
[noissue]


fabricio-aguiar:
Late, but how about ruby bindings?

PR #205, opened 2020-04-06
Print the Pulp config and env vars in the travis before script

mikedep333:
Helps with debugging / using CI.

[noissue]


mikedep333:
Test of it:
https://github.com/pulp/pulpcore/pull/632

PR #207, opened 2020-04-08
Make pulp config acessible to functional tests on travis

mdellweg:
Improve formattig a little.

fixes #6451
https://pulp.plan.io/issues/6451

### DIFF
--- a/templates/travis/.travis/before_script.sh.j2
+++ b/templates/travis/.travis/before_script.sh.j2
@@ -20,7 +20,10 @@ fi
 
 # Developers often want to know the final pulp config
 echo "PULP CONFIG:"
-$CMD_PREFIX bash -c "cat /etc/pulp/settings.*"
+$CMD_PREFIX bash -c "tail -v -n +1 /etc/pulp/settings.*"
+
+# Copy pulp config in a place accessible to the test code
+$CMD_PREFIX bash -c "cat /etc/pulp/settings.py" > $TRAVIS_BUILD_DIR/settings.py
 
 mkdir -p ~/.config/pulp_smash
 
@@ -37,14 +40,14 @@ if [[ "$TEST" == 'pulp' || "$TEST" == 'performance' || "$TEST" == 's3' ]]; then
     $CMD_PREFIX pip3 install pytest mock
     # Many functional tests require these
     $CMD_PREFIX dnf install -yq lsof which dnf-plugins-core
-    {% if docker_fixtures %}
+    {%- if docker_fixtures %}
     # set up pulp-fixtures docker container
     docker run -d -p 0.0.0.0:8000:80 quay.io/pulp/pulp-fixtures:latest
 
     cat ~/.config/pulp_smash/settings.json | \
         jq "setpath([\"custom\",\"fixtures_origin\"]; \"http://$(hostname):8000/fixtures/\")" > temp.json
     cat temp.json > ~/.config/pulp_smash/settings.json
-    {% endif %}
+    {%- endif %}
 fi
 
 if [[ -f $POST_BEFORE_SCRIPT ]]; then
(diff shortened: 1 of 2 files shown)


mdellweg:
@daviddavis This is what i came up. I figured, that the settings.py was generated by the now so called `pulp_installer` and not injected by the ci scripts. So the only way to work on the same config is extracting it from the container again.

daviddavis:
Awesome. This looks good to me. @dkliban or @mikedep333 could you also review?

mdellweg:
The new is used here:
https://github.com/pulp/pulpcore/pull/636

templates/travis/.travis/script.sh.j2
```diff
@@ -13,11 +13,12 @@ export FUNC_TEST_SCRIPT=$TRAVIS_BUILD_DIR/.travis/func_test_script.sh
 # Gets set in .travis/settings.yml, but doesn't seem to inherited by
 # this script.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+export PULP_SETTINGS=$TRAVIS_BUILD_DIR/settings.py
 
 if [ "$TEST" = 'docs' ]; then
-  {% if plugin_name == 'pulpcore' %}
+  {%- if plugin_name == 'pulpcore' %}
     export PULP_CONTENT_ORIGIN=http://$(hostname):24816
```
mdellweg, line 20: Do we still need this with the fix above?

mdellweg: seems to be found here (If `hostname` is the same):
https://travis-ci.com/github/pulp/pulpcore/jobs/316888406#L2846

mikedep333: Are you asking if we need to set PULP_CONTENT_ORIGIN at all? Or if we should limit this to when testing pulpcore?

Which fix are you referring to? A fix that eliminates the need for PULP_CONTENT_ORIGIN? I am not sure what command later in script.sh uses it, the containers are already launched with it set as a config file setting.

templates/travis/.travis/script.sh.j2
```diff
@@ -40,12 +41,15 @@ fi
...
+  ./generate.sh pulp_file python
   pip install ./pulp_file-client
   ./generate.sh pulp_certguard python
-  pip install ./pulp_certguard-client{% else %}  ./generate.sh {{ plugin_snake }} python
+  pip install ./pulp_certguard-client
+{%- else %}
+  ./generate.sh {{ plugin_snake }} python
   pip install ./{{ plugin_snake }}-client
-{% endif %}
+{%- endif %}
```
mikedep333, line 52: Thank you for this cleanup!

templates/travis/.travis/before_script.sh.j2
```diff
@@ -20,7 +20,10 @@ fi
 
 # Developers often want to know the final pulp config
 echo "PULP CONFIG:"
-$CMD_PREFIX bash -c "cat /etc/pulp/settings.*"
+$CMD_PREFIX bash -c "tail -v -n +1 /etc/pulp/settings.*"
+
+# Copy pulp config in a place accessible to the test code
+$CMD_PREFIX bash -c "cat /etc/pulp/settings.py" > $TRAVIS_BUILD_DIR/settings.py
```
mikedep333, line 26: Glad to see this is helping you.

FYI: It is possible that people can change this (they can easily change the k8s cr.yaml that is used to generate settings.py) at runtime via POST_BEFORE_INSTALL, but this is an acceptable risk. We're about to use a single container anyway.

templates/travis/.travis/script.sh.j2
mdellweg: Well for one, why do we need it only for pulpcore? And second with using the `PULP_SETTINGS` variable above (Line 16), we make (among others) the `CONTENT_ORIGIN` setting available to dynaconf. If that is enough and set to the right value already, we could completely skip it here.

templates/travis/.travis/script.sh.j2
mdellweg: you are welcome. (i dislike trailing WS...)

PR #211, opened 2020-04-22
Cleanup README

mdellweg:
* Removed spurious paragraphs
* Removed outdated sections
* Updated completeness checklist

fixes #6474
https://pulp.plan.io/issues/6474

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,6 @@ It's also recommended that you go through the [planning guide](meta_docs/plannin
 The first step is to create a `template_config.yml` for your new plugin. This file contains
 settings used by the `./plugin-template` command when generating new plugins and for future updates.
 
-The first time this file is generated. bootstrap this template. This will create a functional but useless plugin,
-with minimal code, docs, tests, and default Travis CI configuration. Later on we'll discuss
-exactly what each part of this template does and what to change to create a 'real' plugin.
-
 1. Clone this repository
 
    ``$ git clone https://github.com/pulp/plugin_template.git``
@@ -44,7 +40,9 @@ exactly what each part of this template does and what to change to create a 'rea
 
 The first time this command is run, a new directory by the name of PLUGIN_NAME is created inside
 the parent directory of the `plugin_template' directory. The `template_config.yml` is written to
-the root of this new directory. Subsequent uses of the command simply update that file.
+the root of this new directory.
+It is filled with default values for various aspects of the plugin scaffolding.
+You can edit them according to your needs to control subsequent calls to `plugin-template`.
 
 The following settings are stored in `template_config.yml`.
 
@@ -172,6 +170,8 @@ The following settings are stored in `template_config.yml`.
                         This is used during commit validation to make sure the commit is attached to
                         an issue in the correct project.
 
+  pulp_settings         A dictionary of settings that the plugin tests require to be set.
+
 ```
 
 # Bootstrap a new Pulp plugin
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -432,38 +420,6 @@ in the following order, with optional plugin provided scripts in bold:
...
-
-  --travis              Generate or update travis configuration files.
-
-  --docs                Generate or update plugin documentation.
-
-  --all                 Create a new plugin and template all non-excluded files.
-
-  --verbose             Include more output.
-
-```
```
fabricio-aguiar, line 465: why removing it?

mdellweg: Trying to remember, why:
It is outdated, and the information here is doubled anyway. It looked to me like a stop gap until better content was written.

mdellweg:
What is "--all" doing? Is it different from "--bootstrap"? I guess, it at least includes "--bootstrap" which does horrible things on an existing plugin.

fao89:
> What is "--all" doing? Is it different from "--bootstrap"? I guess, it at least includes "--bootstrap" which does horrible things on an existing plugin.

--all executes all args: bootstrap, travis, test,...
as the issue is about out of the box, `--all` is what we have that is more close to it.
as you mentioned about --bootstrap, --all would be terrible for an existing plugin.

README.md
```diff
@@ -191,6 +191,12 @@ functional tests using the [pulp_smash](https://pulp-smash.readthedocs.io/en/lat
 In order to use these tests, you will need to address the "FIXME" messages left in places where
 plugin-writer intervention is required.
 
+At this point, you have a one off chance to use `--all` instead of `--bootstrap` to include all the
+other sections described below.
+
+  **Note** : Regenerating the *bootstrap* section at a later time will reset all files to their
+  original state, which is almost always not intended.
```
fao89, line 198: Nice!

README.md
```diff
@@ -172,6 +170,8 @@ The following settings are stored in `template_config.yml`.
                         This is used during commit validation to make sure the commit is attached to
                         an issue in the correct project.
 
+  pulp_settings         A dictionary of settings that the plugin tests require to be set.
```
fao89, line 173: I didn't know it was not documented yet, thanks!

README.md
melcorr: I think we need to expand upon this a little bit. Can we try and elaborate on what `--all` means here, rather than saying it includes other sections described below? If I use --bootstrap, I get a minimum viable plugin that I can then customize(I'm guessing), if I run `--all`, what specifically do I get? 
The fact that this is an option means that someone might want to use it. Who is this person and what would be their aim? Could we add an example of when it would be an advantage to use `--all` rather than `--bootstrap` at this point, since it is their one chance to do this?

fao89:
```
  --generate-config     Create or update a plugin template config for a plugin and exit.
                        
  --bootstrap           Create a new plugin and template boilerplate code.
                        
  --test                Generate or update functional and unit tests.
                        
  --travis              Generate or update travis configuration files.
                        
  --docs                Generate or update plugin documentation.
                        
```
you can start a plugin interactively:
0 - `./plugin-template --generate-config --plugin-app-label catdog pulp_catdog`
1 - `./plugin-template --generate-config --plugin-app-label catdog pulp_catdog`
with the previous steps, you can have a minimum viable plugin, and choose whether you add tests, CI or docs.
**Note:** step 0 is mandatory, as plugin_template depends on the config
So, if a week later I want to add tests I can:
2 - `./plugin-template --test pulp_catdog`

If the plugin writer wants a full plugin since the beginning, `--all` would be useful as it executes (bootstrap, test, travis, docs). In our CI we want to build a full plugin for testing, so we use `--all`:
https://github.com/pulp/plugin_template/blob/master/.travis/bootstrap_catdog.sh

melcorr: How about something like: 

> At this point, you have a one-off opportunity to use the --all option, which generates everything included in the --bootstrap option, as well as documentation, functional and unit test, and travis configuration file templates that you require to support a plugin.

quba42:
Just saw this PR. Is this ready to go? (It might be better for me to work on https://github.com/pulp/plugin_template/pull/272 on top of this change.)

mdellweg:
I'd be happy to merge it right away.
Ping for awareness: @fao89

fao89:
> I'd be happy to merge it right away.
> Ping for awareness: @fao89

I've been waiting for this, please merge it!

PR #213, opened 2020-04-23
Adding script to check for non-plugin pulpcore imports

daviddavis:
fixes #4894

### DIFF
--- a/templates/bootstrap/plugin_name/app/viewsets.py.j2
+++ b/templates/bootstrap/plugin_name/app/viewsets.py.j2
@@ -11,7 +11,7 @@ from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
-from pulpcore.app.viewsets import RemoteFilter
+from pulpcore.plugin.viewsets import RemoteFilter
 from pulpcore.plugin import viewsets as core
 from pulpcore.plugin.actions import ModifyRepositoryActionMixin
 from pulpcore.plugin.serializers import (
--- a/templates/travis/.travis/check_pulpcore_imports.sh.j2
+++ b/templates/travis/.travis/check_pulpcore_imports.sh.j2
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+{% include 'header.j2' %}
+
+set -uv
+
+# check for imports not from pulpcore.plugin. exclude tests
+MATCHES=$(grep -n -r --include \*.py "from pulpcore.*import" . | grep -v "tests\|plugin")
+
+if [ $? -ne 1 ]; then
+  printf "\nERROR: Detected bad imports from pulpcore:\n"
+  echo "$MATCHES"
+  printf "\nPlugins should import from pulpcore.plugin."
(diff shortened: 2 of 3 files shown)


templates/travis/.travis/check_pulpcore_imports.sh.j2
```diff
@@ -0,0 +1,14 @@
...
+set -euv
+
+# exlude tests and imports from pulpcore.plugin
+grep "from pulpcore" **/*.py | grep -v "tests\|plugin"
+
+if [ $? -eq 1]; then
+  exit 0
+else
+  exit 1
+fi
```
daviddavis, line 14: My bash fu is weak :cry:

templates/bootstrap/plugin_name/app/viewsets.py.j2
```diff
@@ -11,7 +11,7 @@ from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
-from pulpcore.app.viewsets import RemoteFilter
+from pulpcore.plugin.viewsets import RemoteFilter
```
daviddavis, line 14: :smile:

daviddavis: Looks like it's missing from the plugin API. Fix: https://github.com/pulp/pulpcore/pull/666

templates/travis/.travis/check_pulpcore_imports.sh.j2
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+{% include 'header.j2' %}
+
+set -uv
+
+# check for imports not from pulpcore.plugin. exclude tests
+MATCHES=$(grep -r --include \*.py "from pulpcore.*import" . | grep -v "tests\|plugin")
```
mdellweg, line 8: I suspect this command to present all findings collapsed in a single line.
If you add a `-n` to the first `grep`, it should report the line-number next to the filename, where it found the line.

templates/travis/.travis/check_pulpcore_imports.sh.j2
```diff
@@ -0,0 +1,16 @@
...
+
+if [ $? -eq 1 ]; then
```
mdellweg, line 10: If you invert that expression, you would not need an extra branch for `exit 0`.

templates/travis/.travis/check_pulpcore_imports.sh.j2
daviddavis: Will add `-n`. It does seem to report $MATCHES on newlines when I use echo double quotes:

```
ERROR: Detected bad imports from pulpcore:
./pulp_rpm/app/modulemd.py:4:from pulpcore.app.models.content import Artifact
./pulp_rpm/app/modelresource.py:1:from pulpcore.app.modelresource import QueryModelResource

Plugins should import from pulpcore.plugin.                                                                                                                                                                                    
```

PR #214, opened 2020-04-24
Add check for changelog extension

daviddavis:
fixes #6571

### DIFF
--- a/templates/travis/.travis/validate_commit_message.py.j2
+++ b/templates/travis/.travis/validate_commit_message.py.j2
@@ -1,6 +1,7 @@
 {% include 'header.j2' %}
 
 import glob
+import os
 import re
 import requests
 import subprocess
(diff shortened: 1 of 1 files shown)


templates/travis/.travis/validate_commit_message.py.j2
```diff
@@ -27,8 +28,14 @@ def __check_status(issue):
 
 
 def __check_changelog(issue):
-    if len(glob.glob(f"CHANGES/**/{issue}.*", recursive=True)) < 1:
+    matches = glob.glob(f"CHANGES/**/{issue}.*", recursive=True)
+
+    if len(matches) < 1:
         sys.exit(f"Could not find changelog entry in CHANGES/ for {issue}.")
+    for match in matches:
+        regex = re.compile('CHANGES/.*\.({})$'.format(("|").join(CHANGELOG_EXTS)))
```
mdellweg, line 36: You should be able to just check, whether the result of `path.splitext` is in `CHANGELOG_EXTS`.

https://docs.python.org/3/library/os.path.html#os.path.splitext

PR #215, opened 2020-05-05
Introducing release script

fao89:
which executes the first 6 steps of https://pulp.plan.io/projects/pulp/wiki/Pulp3_Release_Guide

https://pulp.plan.io/issues/6600
closes #6600

```
python .travis/release.py new_version new_dev_version current_pulpcore_version next_pulpcore_version
```

Executing the script for `pulp_file`:
```
python .travis/release.py 1.0.0 1.0.1 4.0 4.1
```
Output:
https://github.com/pulp/pulp_file/compare/master...fao89:release_1.0.0

Executing the script for `pulpcore`:
```
python .travis/release.py 3.3.1 3.4.0
```
Output:
https://github.com/pulp/pulpcore/compare/master...fao89:release_3.3.1

### DIFF
--- a/templates/travis/.travis/release.py.j2
+++ b/templates/travis/.travis/release.py.j2
@@ -1,0 +1,131 @@
+import argparse
+import os
+import textwrap
+
+from git import Repo
+
+
+REDMINE_QUERY_URL = "https://pulp.plan.io/issues?set_filter=1&status_id=*&issue_id="
+release_path = os.path.dirname(os.path.abspath(__file__))
+plugin_path = release_path
+if ".travis" in release_path:
+    plugin_path = os.path.dirname(release_path)
+
+version = {}
+with open(f"{plugin_path}/{{ plugin_name }}/__init__.py") as fp:
+    version_line = [line for line in fp.readlines() if "__version__" in line][0]
+    exec(version_line, version)
+release_version = version["__version__"].replace(".dev", "")
+
+to_close = []
+for filename in os.listdir(f"{plugin_path}/CHANGES"):
+    if filename.split(".")[0].isdigit():
+        to_close.append(filename.split(".")[0])
+issues = ",".join(to_close)
+
+helper = textwrap.dedent(
+    """\
+        Start the release process.
+
+        Example:
+            setup.py on plugin before script:
+                version="2.0.dev"
+                requirements = ["pulpcore>=3.4"]
+
+
+            $ python .travis/realease.py minor 4.0 4.1
+
+            setup.py on plugin after script:
+                version="2.1.dev"
+                requirements = ["pulpcore>=4.0,<4.1"]
+
+    """
+)
+parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter, description=helper)
+
+parser.add_argument(
+    "release_part", type=str, help="Whether the release should be major, minor or patch.",
+)
+
+parser.add_argument(
+    "--lower", type=str, required=False, help="Lower bound of pulpcore requirement.",
+)
+
+parser.add_argument(
+    "--upper", type=str, required=False, help="Upper bound of pulpcore requirement.",
+)
+
+args = parser.parse_args()
+
+release_part = args.release_part
+
+if "pulpcore" not in release_path:
+    lower_pulpcore_version = args.lower
+    upper_pulpcore_version = args.upper
+
+print("\n\nHave you checked the output of: $towncrier --version x.y.z --draft")
+print(f"\n\nRepo path: {plugin_path}")
+repo = Repo(plugin_path)
+git = repo.git
+
+git.checkout("HEAD", b=f"release_{release_version}")
+
+# First commit: changelog
+os.system(f"towncrier --yes --version {release_version}")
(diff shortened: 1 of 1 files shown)


templates/travis/.travis/release.sh.j2
```diff
@@ -0,0 +1,58 @@
...
+git add .
+git commit -F- <<EOF
+Building changelog for $VERSION
+
+[noissue]
+EOF
+
+# Second commit: release version
+sed -i "s/\.dev//g" setup.py
+# Missing sed on pulpcore requirement on setup.py
```
fao89, line 36: I could not think of a good sed for this

mdellweg: maybe something like
`sed -i -e '/version=/s/\.dev//'
To only perform it once and on a line that contains "version=".

fao89: Sorry, I did not describe well the problem, what I need to do is find a way to:
`pulpcore>=3.3` into this `pulpcore>=3.4,<3.5`

templates/travis/.travis/release.sh.j2
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+export PLUGIN_DIR={{ plugin_name }}
+export BASEDIR=$(dirname $(readlink -f "$0"))
+export NEW_VERSION=$1
+export DEV_VERSION=$VERSION\.dev
+
+cd $BASEDIR/..
+
+towncrier --version $NEW_VERSION --draft
```
daviddavis, line 13: I wouldn't bother with this. Once we fully automate this and move it to a CI, we won't be able to run this step.

I think the release nanny will have to run this before triggering the release.

daviddavis:
The logic looks great. Very glad to see this.

fao89:
I remembered why I did not use pygithub, it was about credentials:
https://github.com/pulp/pulp_file/blob/master/.travis/cherrypick.py#L9-L12
Running it locally would require to set GITHUB_TOKEN for example, and with bash script you can just use git commands

templates/travis/.travis/release.sh.j2
mdellweg: I agree. Those changelog entries probably need to be revisited redactorially anyway.

fao89:
### Executing the script:
```
python .travis/release.py 1.0.1 4.0 4.1
```
### Output:
https://github.com/pulp/pulp_file/compare/master...fao89:release_1.0.0

templates/travis/.travis/release.py.j2
```diff
@@ -0,0 +1,122 @@
...
+print("\n\nHave you checked the output of: $towncrier --version x.y.z --draft")
+print(f"\n\nRepo path: {plugin_path}")
+repo = Repo(plugin_path)
+git = repo.git
+
+git.checkout("HEAD", b=f"release_{new_version}")
+
+# First commit: changelog
+os.system(f"towncrier --yes --version {new_version}")
+git.add(".")
```
daviddavis, line 59: Might just want to check in CHANGES.rst and anything in CHANGES/

templates/travis/.travis/release.py.j2
```diff
@@ -0,0 +1,125 @@
+"""
+Release script
+
+$ python .travis/realease.py
```
daviddavis, line 4: nitpick - s/realease/release/

templates/travis/.travis/release.py.j2
```diff
@@ -0,0 +1,125 @@
...
+    for line in setup_lines:
+        if "version=" in line:
+            line = line.replace(current_dev_version, new_version)
+        if "pulpcore" in line and "entry_points" not in line and "pulpcore" not in release_path:
+            sep = "'" if len(line.split('"')) == 1 else '"'
+            for word in line.split(sep):
+                if "pulpcore" in word:
+                    pulpcore_word = word
+
+            line = line.replace(
```
bmbouter, line 77: Can we use bump2version to do this instead? https://pypi.org/project/bump2version/ Since that's software is made to do that I think it would be better.

That is also one of the recommended tools to help "single source" python versions https://packaging.python.org/guides/single-sourcing-package-version/ It would be cool if the plugin_template configured plugins with bump2version.

fao89: I'm reading the docs, but so far I could not understand how to deal with `.dev` using `bump2version`

bmbouter: What about `--new-version 0.1.2.dev` ?

fao89: I believe I understood how to do it, but I would need to open PR against all plugins first, I will do it on pulp_file as PoC

fao89: draft PR on `pulp_file`:
https://github.com/pulp/pulp_file/pull/390

```
python .travis/release.py major 4.0 4.1
```

```
Args:
    release_part - major/minor/patch
    current_pulpcore_version - lower bound of pulpcore requirement
    next_pulpcore_version - upper bound of pulpcore requirement
```

PR #217, opened 2020-05-12
Add option to disable import check

daviddavis:
[noissue]


daviddavis:
FYI, the migration plugin requested this but it sounds like they may not need it. So waiting to hear back from them.

daviddavis:
Closing out per open floor today.

PR #219, opened 2020-05-13
Parametrize pulp URL in docs Makefile

quba42:
[noissue]

This ensures that the docs can be built both from within pulplift source
boxes, as well as the Travis pipeline.

The idea came from here: https://github.com/pulp/pulp_deb/pull/166#discussion_r420258701
The implementation came from @mdellweg 

I have successfully tested it on the pulp_deb plugin.


quba42:
The failed checks appear to relate to the Travis infrastructure, and not the change:
```
fatal: unable to access 'https://github.com/pulp/plugin_template.git/': The requested URL returned error: 429
```

PR #220, opened 2020-05-14
Change nginx.conf to match with pulp_installer one

fao89:
https://github.com/pulp/pulp_installer/blob/master/roles/pulp-webserver/templates/nginx.conf.j2
[noissue]

### DIFF
--- a/templates/travis/.travis/assets/nginx.conf
+++ b/templates/travis/.travis/assets/nginx.conf
@@ -41,9 +41,6 @@ http {
         # Gunicorn docs suggest this value.
         keepalive_timeout 5;
 
-        # Path to Pulp's static files.
-        root /usr/local/lib/pulp/lib/python3.7/site-packages/rest_framework;
-
         location /pulp/content/ {
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
             proxy_set_header X-Forwarded-Proto $scheme;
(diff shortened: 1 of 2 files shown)


templates/travis/.travis/assets/nginx.conf
```diff
@@ -41,17 +41,14 @@ http {
...
-
         location /pulp/content/ {
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
             proxy_set_header X-Forwarded-Proto $scheme;
             proxy_set_header Host $http_host;
             # we don't want nginx trying to do something clever with
             # redirects, we set the Host: header above already.
             proxy_redirect off;
-            proxy_pass http://127.0.0.1:24816;
+            proxy_pass http://pulp-content;
```
mdellweg, line 51: Where does this change come from? I am not sure it fits the single container architecture.

fao89: it comes from here:
https://github.com/pulp/plugin_template/blob/master/templates/travis/.travis/assets/nginx.conf#L25-L27
and this change was applied to some plugins:
- https://github.com/pulp/pulp_rpm/blob/master/.travis/assets/nginx.conf
- https://github.com/pulp/pulp_ansible/blob/master/.travis/assets/nginx.conf

mdellweg: Thank you for clarifying. I did no see this. And nginx config always seems to be good for a surprise.

fao89:
@dkliban @daviddavis could you merge it? I could not due to travis issue

PR #224, opened 2020-05-28
Switch to using fixtures from docker.io

daviddavis:
fixes #6767


templates/travis/.travis/install.sh.j2
```diff
@@ -131,7 +131,7 @@ VARSYAML
 if [[ "$TEST" == "pulp" || "$TEST" == "performance" || "$TEST" == "s3" ]]; then
   sed -i -e '/^services:/a \
   - name: pulp-fixtures\
-    image: quay.io/pulp/pulp-fixtures:latest' vars/main.yaml
+    image: docker.io/pulp/pulp-fixtures:latest' vars/main.yaml
```
fao89, line 134: :heart:

PR #226, opened 2020-06-02
Add nocoverage on release script

fao89:
[noissue]


daviddavis:
I'm hesitant to make plugin specific changes to the release script. What if we update pulp_rpm's coverage check to be smarter? Perhaps it could look for actual code changes (ie changes to .py files in pulp_rpm/app)?

fao89:
> I'm hesitant to make plugin specific changes to the release script. What if we update pulp_rpm's coverage check to be smarter? Perhaps it could look for actual code changes (ie changes to .py files in pulp_rpm/app)?

I agree with you, but in this case, I thought adding `[nocoverage]` to commit message was not a meaningful change to the other plugins.

fao89:
Closing in favor of https://github.com/pulp/pulp_rpm/pull/1734

PR #228, opened 2020-06-03
Update validate commit script to use pathlib

daviddavis:
[noissue]


templates/travis/.travis/validate_commit_message.py.j2
```diff
@@ -37,12 +37,12 @@ def __check_status(issue):
 
 
 def __check_changelog(issue):
-    matches = glob.glob(f"CHANGES/**/{issue}.*", recursive=True)
+    matches = list(Path("CHANGES").rglob(f"{issue}.*"))
```
fao89, line 40: Nice!

PR #236, opened 2020-06-16
Fixing a missing slash for fixtures container

daviddavis:


daviddavis:
Looks like pulp_rpm needs this.

PR #238, opened 2020-06-30
Fix release script for certguard

fao89:
[noissue]


fao89:
Ref: https://github.com/pulp/pulp-certguard/pull/70

PR #240, opened 2020-07-03
Validate redmine milestone and close issues

fao89:
https://pulp.plan.io/issues/6899
closes #6899

### DIFF
--- a/templates/travis/.travis/redmine.py.j2
+++ b/templates/travis/.travis/redmine.py.j2
@@ -1,0 +1,25 @@
+{% include 'header.j2' %}
+
+import os
+import sys
+
+from redminelib import Redmine
+
+REDMINE_API_KEY = os.environ["REDMINE_API_KEY"]
+REDMINE_QUERY_URL = sys.argv[1]
+CLOSED_CURRENTRELEASE = 11
+
+redmine = Redmine(REDMINE_QUERY_URL.split("issues")[0], key=REDMINE_API_KEY)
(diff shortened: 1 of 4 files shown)


templates/travis/.travis/redmine.py.j2
```diff
@@ -0,0 +1,20 @@
...
+query_issues = REDMINE_QUERY_URL.split("=")[-1].split(",")
+
+for issue in query_issues:
+    status = redmine.issue.get(int(issue)).status.name
+    if status != "MODIFIED":
+        raise ValueError("One or more issues are not MODIFIED")
+
+for issue in query_issues:
+    print(f"Closing #{issue}")
+    redmine.issue.update(int(issue), status_id=CLOSED_CURRENTRELEASE)
```
daviddavis, line 20: Why is this file a jinja file?

mdellweg: It contains `{% include 'header.j2' %}`

fao89: I didn't put the header, I copied from https://github.com/pulp/pulp_file/pull/409
and automatically changed the python files into Jinja ones without thinking.
Now I'm not sure if we need to add the header or change it back to python file

daviddavis: Where?

daviddavis: Including the header is fine. I added the header but kinda regret doing so because I can't use python syntax highlighting when viewing these files.

PR #243, opened 2020-08-03
Making validate milestone optional

fao89:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -48,6 +48,7 @@ DEFAULT_SETTINGS = {
     'stable_branch': None,
     'travis_addtl_services': [],
     'travis_notifications': None,
+    'update_redmine': False,
 }
 
 
--- a/templates/travis/.travis.yml.j2
+++ b/templates/travis/.travis.yml.j2
@@ -131,6 +131,14 @@ jobs:
       if: type = cron
       {%- endif %}
     {%- endif %}
+    {%- if update_redmine %}
+    - stage: update-redmine
+      before_install: skip
+      install: skip
+      before_script: skip
+      script: bash .travis/update_redmine.sh
+      if: tag IS present
+    {%- endif %}
     {%- if cherry_pick_automation %}
     - stage: cherry-pick
       if: type = cron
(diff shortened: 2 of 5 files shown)


fao89:
@goosemania could you try it and see if it solves the problem?

fao89:
@goosemania FYI: all changes came from https://pulp.plan.io/issues/6899

plugin-template
```diff
@@ -48,6 +48,7 @@ DEFAULT_SETTINGS = {
     'stable_branch': None,
     'travis_addtl_services': [],
     'travis_notifications': None,
+    'update_redmine': False,
```
goosemania, line 51: Does it makes sense to set it to True by default? Just because it's what it does right now.

templates/travis/.travis.yml.j2
```diff
@@ -131,6 +131,14 @@ jobs:
       if: type = cron
       {%- endif %}
     {%- endif %}
+    {%- if update_redmine %}
+    - stage: update-redmine
+      if: tag IS present
```
goosemania, line 136: For consistency, maybe it's worth having the condition being the last.

templates/travis/.travis/release.py.j2
```diff
@@ -53,7 +60,7 @@ def validate_redmine_data(redmine_query_url, redmine_issues):
         error_messages.append(f"Verify at {redmine_query_url}")
         raise RuntimeError("\n".join(error_messages))
 
-    return f"{REDMINE_URL}/versions/{milestone_id}.json"
+    return commit_message
```
goosemania, line 63: For readability, I suggest to settle on one name and not use `commit_message` and `milestone_url` for the same data.

plugin-template
fao89: With the errors you faced, I believe it is better to start as False, so the plugin developer is consciously enabling it

goosemania: @fao89, @dkliban , ok, however then the rollout of this change should flip it to True in the majority of the cases.

PR #245, opened 2020-08-10
Starting to check commits

fao89:


.travis/bootstrap_catdog.sh
```diff
@@ -20,4 +22,4 @@ cp ../plugin_template/.travis/post_before_script.sh .travis/
 # generate a git commit
 git init
 git add .
-git commit -m $'Initial commit\n\n[noissue]'
+git commit -m "${COMMIT_MSG}"
```
daviddavis, line 25: It feels like a hack having the plugin bootstrap test check plugin_template's commit messages. For one thing, I am guessing it doesn't handle multiple commits?

What if we convert templates/travis/.travis/validate_commit_message.py.j2 to templates/travis/.travis/validate_commit_message.py (the only thing we use jinja for in this file is to add a header which I don't think we need) and then write a small github workflow or travis job to run it?

fao89: it is for doing the `Required PR` thing, and it only checks the last commit

daviddavis: Don't we want to validate all commits?

daviddavis: It also still feels weird to me that we're using the bootstrap test code to validate commit messages.

README.md
```diff
@@ -487,3 +487,15 @@ optional arguments:
...
 - [ ] Errors are handled according to Pulp conventions
 - [ ] Docs for plugin are available (any location and format preferred and provided by plugin writer)
+
+
+# Contributing
+
+### Pull Request Checklist
+
+1. Unless a change is small or doesn't affect users, create an issue on
+https://pulp.plan.io/projects/pulp . Set the Category to "Installer".
```
daviddavis, line 497: s/affect uesrs/affect plugin writers/
s/Set the Category to "Installer"//Add the tag "Plugin Template"./

Also remove the unnecessary space before the period.

PR #248, opened 2020-08-11
Some tweaks to the commit checking

daviddavis:
I overlooked [another line that uses jinja](https://github.com/pulp/plugin_template/blob/92f12e455d26ec9c863ec0c9e9d678bf6714bc6d/templates/travis/.travis/validate_commit_message.py#L14) so I switched the validate script back to j2 and am using the j2cli tool to generate the file.

### DIFF
--- a/.github/workflows/check_commits.yaml
+++ b/.github/workflows/check_commits.yaml
@@ -14,14 +14,20 @@ jobs:
           python-version: 3.8
       - name: Install python dependencies
         run: |
-          pip install requests
+          pip install requests j2cli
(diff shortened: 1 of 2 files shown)


.github/workflows/check_commits.yaml
```diff
@@ -14,14 +14,20 @@ jobs:
           python-version: 3.8
       - name: Install python dependencies
         run: |
-          pip install requests
+          pip install requests j2cli
```
fao89, line 17: Nice!

PR #251, opened 2020-08-11
Fixed the UnboundLocalError when no issues are assigned to a milestone

goosemania:
The full error was "UnboundLocalError: local variable 'milestone_id'
referenced before assignment".

[noissue]

### DIFF
--- a/templates/travis/.travis/release.py.j2
+++ b/templates/travis/.travis/release.py.j2
@@ -36,15 +36,19 @@ def validate_redmine_data(redmine_query_url, redmine_issues):
             stats["status_not_modified"].append(issue)
 
         {%- if update_redmine %}
+        milestone_id = None
         try:
             milestone = redmine_issue.fixed_version.name
             milestone_id = redmine_issue.fixed_version.id
             stats[f"milestone_{milestone}"].append(issue)
         except ResourceAttrError:
             stats["without_milestone"].append(issue)
-
-        milestone_url = f"RedmineMilestone: {REDMINE_URL}/versions/{milestone_id}.json\n[noissue]"
         {%- endif %}
+
+    {%- if update_redmine %}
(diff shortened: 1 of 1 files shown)


templates/travis/.travis/release.py.j2
```diff
@@ -42,9 +43,12 @@ def validate_redmine_data(redmine_query_url, redmine_issues):
             stats[f"milestone_{milestone}"].append(issue)
         except ResourceAttrError:
             stats["without_milestone"].append(issue)
+        {%- endif %}
 
+    {%- if update_redmine %}
+    if milestone_id is not None:
```
fao89, line 49: it is weird, the milestone should be set on line 42

fao89: it should have a milestone_id, otherwise, the job for updating redmine will fail

goosemania: if there are no issues assigned to a milestone, it won't be set.

goosemania: the script will stop early if there is at least one issue not assigned to a milestone, it won't proceed further.

fao89: so we need an if here:
https://github.com/pulp/plugin_template/blob/master/templates/travis/.travis/update_redmine.sh
to check it the commit message contains the redmine query:
https://github.com/pulp/plugin_template/blob/master/templates/travis/.travis/update_redmine.sh

goosemania: What is the case when there is no redmine query and validate method passes so it gets to the update_redmine.sh?
A release with no redmine issues is the only case I can think of.

fao89: I believe this is the only case

PR #253, opened 2020-08-26
Adds support for publishing docs to pulpproject.org

dkliban:
re: #7229
https://pulp.plan.io/issues/7229

### DIFF
--- a/templates/travis/.travis/docs-builder.py.j2
+++ b/templates/travis/.travis/docs-builder.py.j2
@@ -1,0 +1,169 @@
+#!/usr/bin/env python
+
+import argparse
+import subprocess
+import os
+import re
+from shutil import rmtree
+import tempfile
+
+WORKING_DIR = os.environ["TRAVIS_BUILD_DIR"]
+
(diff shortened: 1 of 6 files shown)


templates/travis/.travis/docs-builder.py.j2
```diff
@@ -0,0 +1,171 @@
...
+    ga_build = False
+
+    if not re.search("[a-zA-Z]", branch) and len(branch.split(".")) > 2:
+        ga_build = True
+
+    # build the docs via the Pulp project itself
+    print("Building the docs")
+    docs_directory = os.sep.join([WORKING_DIR, "docs"])
+
+    make_command = ["make", "PULP_URL=http://pulp", "diagrams", "html"]
```
fao89, line 89: Not about this PR, but I think we should have `make bindings_docs`

dkliban: yes, we will add that separately.

daviddavis:
Code LGTM. What about some docs in the README and also a changelog entry?

daviddavis (commented):
Thanks for adding a changelog and README info. Looks great.

daviddavis (approved):

PR #254, opened 2020-08-27
Communicate the black version running in pipelines

quba42:
Just a thought I had to give people some indication why their local black might produce different results than the pipeline's black.

[noissue]

### DIFF
--- a/templates/travis/.travis/before_install.sh.j2
+++ b/templates/travis/.travis/before_install.sh.j2
@@ -46,6 +46,7 @@ pip install -r dev_requirements.txt
 
 {% if black -%}
 # run black separately from flake8 to get a diff
+black --version
 black --check --diff .
 {%- endif %}
 


quba42:
And the test result is a case in point why this is needed!

mdellweg (approved):
I like the idea.

quba42:
Ugh! This looks like it is generating scripts from template and then running black on them.
The changes need to find there way back into the right template. How tedious.

quba42:
Or I am still getting different results from the pipeline, even when using the same black version...

daviddavis:
Here's the fix for the black issues: https://github.com/pulp/plugin_template/pull/255

The changes to cherrypick.py are unnecessary. Can you revert them?

templates/travis/.travis/before_install.sh.j2
```diff
@@ -46,6 +46,7 @@ pip install -r dev_requirements.txt
 
 {% if black -%}
 # run black separately from flake8 to get a diff
+black --version
```
fao89, line 49: sweet!

quba42:
I rebased, so the tests _should_ turn green now.

PR #255, opened 2020-08-27
Fix black issue with new black version

daviddavis:
[noissue]


quba42 (approved):
The pipeline is green, so this appears to be what the new black version wants.
Since it is all just reformatting, not sure how much more review it needs.

PR #256, opened 2020-08-27
Have commit validation accept different cases for keywords

daviddavis:
fixes #7404

### DIFF
--- a/templates/travis/.travis/validate_commit_message.py.j2
+++ b/templates/travis/.travis/validate_commit_message.py.j2
@@ -49,7 +49,7 @@ print("Checking commit message for {sha}.".format(sha=sha[0:7]))
 
 # validate the issue attached to the commit
 regex = r"(?:{keywords})[\s:]+#(\d+)".format(keywords=("|").join(KEYWORDS))
-pattern = re.compile(regex)
+pattern = re.compile(regex, re.IGNORECASE)
 
 issues = pattern.findall(message)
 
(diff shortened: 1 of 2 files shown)


templates/travis/.travis/validate_commit_message.py.j2
```diff
@@ -51,7 +51,7 @@ print("Checking commit message for {sha}.".format(sha=sha[0:7]))
 regex = r"(?:{keywords})[\s:]+#(\d+)".format(keywords=("|").join(KEYWORDS))
 pattern = re.compile(regex)
 
-issues = pattern.findall(message)
+issues = pattern.findall(message, flags=re.IGNORECASE)
```
fao89, line 54: GHA did not like it:
```
TypeError: 'flags' is an invalid keyword argument for findall()
```
https://github.com/pulp/plugin_template/pull/256/checks?check_run_id=1038033183#step:7:23

dralley:
@daviddavis Any blocker to merge?

mdellweg:
I think, the only question was, whether the communication with pulp.plan.io was also working with capitalized keywords.
If that is still true, we could add a `tolower` there.

daviddavis:
Looks like based on [this test](https://www.redmine.org/projects/redmine/repository/revisions/2637/entry/trunk/test/unit/changeset_test.rb#L26) that keyword checking is case insensitive for redmine. Merging.

PR #257, opened 2020-08-29
Added support for deprecation type to towncrier

daviddavis:
fixes #7421

### DIFF
--- a/templates/bootstrap/pyproject.toml.j2
+++ b/templates/bootstrap/pyproject.toml.j2
@@ -13,6 +13,36 @@ title_format = "{version} ({project_date})"
 template = "CHANGES/.TEMPLATE.rst"
 issue_format = "`#{issue} <https://pulp.plan.io/issues/{issue}>`_"
 
+    [[tool.towncrier.type]]
+        directory = "feature"
+        name = "Features"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "bugfix"
+        name = "Bugfixes"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "doc"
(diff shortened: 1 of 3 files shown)


templates/bootstrap/pyproject.toml.j2
```diff
@@ -13,6 +13,11 @@ title_format = "{version} ({project_date})"
 template = "CHANGES/.TEMPLATE.rst"
 issue_format = "`#{issue} <https://pulp.plan.io/issues/{issue}>`_"
 
+    [[tool.towncrier.type]]
+        directory = "deprecation"
+        name = "Deprecations"
+        showcontent = true
+
```
mdellweg, line 20: Now i am wondering if this file should still be in the "bootstrap" section, because that makes it a mess to update in existing plugins.

fao89: when I created a new type for pulp_installer, it stopped to recognize the old ones, I had to do it:
https://github.com/pulp/pulp_installer/blob/master/pyproject.toml#L36-L39
(Maybe I missed something there, I'm just pointing it out, for us to test this change)

daviddavis: @fao89 I tested it out and you were right.

@mdellweg I'm a bit mixed since I agree that making this change to existing plugins will be a pain but also, I wonder if pyproject.toml will diverge across projects.

mdellweg: This question (moving files) does not need to be answered here, obviously.

daviddavis: @mdellweg it's definitely worth a discussion.

PR #261, opened 2020-09-07
Add towncrier to release requirements

quba42:
[noissue]

If I am running the release script locally in a python venv, then I need `towncrier` in it.

I am not sure if there is an issue with having this requirement in both `release_requirements.txt` and `doc_requirements.txt` (or what effect this has on Travis pipelines).

What do you think?

### DIFF
--- a/templates/travis/.travis/release_requirements.txt.j2
+++ b/templates/travis/.travis/release_requirements.txt.j2
@@ -1,3 +1,4 @@
 bump2version
 gitpython
 python-redmine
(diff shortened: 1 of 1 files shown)


mdellweg:
I think, it is only a problem to have the same dependency twice in one pip run. Despite of that, pips dependency management is a mess.
OTOH, the travis pipeline runs on this repo for a freshly generated plugin.

mdellweg (approved):

PR #263, opened 2020-09-10
Use pip 2020 resolver

fao89:
[noissue]


mdellweg:
Is this a precaution, or do we have an issue this is supposed to solve?
(I realized it wasn't the solution for our latest installer CI failure)

daviddavis:
I believe this sort of change should be documented in the changelog as well.

fao89:
it is for matching with pulp-oci-images:
https://github.com/pulp/pulp-oci-images/blob/master/pulp_fedora31/Containerfile#L11

mdellweg:
As far as i understand this activates a feature in preview state. Not sure we want this without a reason. OTOH maybe we want it to know we are compatible with the next version of pip.

PR #264, opened 2020-09-11
Remove unused services from travis config

daviddavis:
fixes #7500

### DIFF
--- a/CHANGES/7500.misc
+++ b/CHANGES/7500.misc
@@ -1,0 +1,1 @@
+Removing some unused services from travis config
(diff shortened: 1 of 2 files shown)


CHANGES/7500.bugfix
```diff
@@ -0,0 +1 @@
+Removing some unused services from travis config
```
dralley, line 1: It doesn't really matter but I would mark this as "misc" rather than "bugfix".

Don't feel obligated to change it though.

dralley (approved):

PR #265, opened 2020-09-18
Fixes bug with the docs building stage conditions for the continuous …

dkliban:
…docs.

Also fixes some flake 8 issues that were needed for the pulp_ansible repo.

[noissue]


daviddavis (approved):

fao89 (approved):
Thank you!

PR #266, opened 2020-09-18
Add diagrams at Makefile

fao89:
[noissue]

### DIFF
--- a/templates/docs/docs/Makefile.j2
+++ b/templates/docs/docs/Makefile.j2
@@ -2,11 +2,13 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W -n
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = _build
-PULP_URL      = "http://localhost:24817"
+SPHINXOPTS    		  = -W -n
+SPHINXBUILD   		  = sphinx-build
+PAPER         		  =
+BUILDDIR      		  = _build
+DIAGRAM_BUILD_DIR     = _diagrams
+DIAGRAM_SOURCE_DIR    = diagrams_src
+PULP_URL      		  = "http://localhost:24817"
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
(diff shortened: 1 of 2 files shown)


templates/docs/docs/Makefile.j2
```diff
@@ -2,11 +2,13 @@
...
-PAPER         =
-BUILDDIR      = _build
-PULP_URL      = "http://localhost:24817"
+SPHINXOPTS    		  = -W -n
+SPHINXBUILD   		  = sphinx-build
+PAPER         		  =
+BUILDDIR      		  = _build
+DIAGRAM_BUILD_DIR     = _diagrams
+DIAGRAM_SOURCE_DIR    = diagrams_src
+PULP_URL      		  = "http://localhost:24817"
```
mdellweg, line 11: Somehow, the alignment of the equals went wrong here? Is this some uncommon tab expansion?
I would not even bother to align them, as in "Why do i need to touch all the lines if one variable name becomes longer?"

templates/docs/docs/Makefile.j2
```diff
@@ -72,7 +85,7 @@ htmlhelp:
 	$(SPHINXBUILD) -b htmlhelp $(ALLSPHINXOPTS) $(BUILDDIR)/htmlhelp
 	@echo
 	@echo "Build finished; now you can run HTML Help Workshop with the" \
-	      ".hhp project file in $(BUILDDIR)/htmlhelp."
+		  ".hhp project file in $(BUILDDIR)/htmlhelp."
```
mdellweg, line 88: Also some formatting strangeness around tabs?

mdellweg (changes requested):
Would you mind explaining the reason for this change, as it goes as [noissue] (and is not obvious to me)?

fao89:
> Would you mind explaining the reason for this change, as it goes as [noissue] (and is not obvious to me)?

I forgot to add it here, it was a fix for pulp_file,
PR: https://github.com/pulp/pulp_file/pull/431
issue: https://travis-ci.com/github/pulp/pulp_file/jobs/387174826#L1220

templates/docs/docs/Makefile.j2
fao89: this was looking different on my editor, I think it was a space/tab issue with my editor config

templates/docs/docs/Makefile.j2
fao89: his was looking different on my editor, I think it was a space/tab issue with my editor config

templates/docs/docs/Makefile.j2
fao89: I checked it at vscode, sublime, and vim. And I couldn't reproduce it locally
image

mdellweg (approved):

dkliban (approved):

PR #267, opened 2020-09-22
Adds a stage to test released plugin with pulpcore's master branch.

dkliban:
re: #7411
https://pulp.plan.io/issues/7411

### DIFF
--- a/templates/travis/.travis.yml.j2
+++ b/templates/travis/.travis.yml.j2
@@ -143,6 +143,12 @@ jobs:
       before_script: skip
       script: bash .travis/cherrypick.sh
     {%- endif %}
+    {%- if test_released_plugin_with_next_pulpcore_release %}
+    - stage: test_released_plugin_with_next_pulpcore_release
+      env:
+        - TEST=plugin-from-pypi
+      if: type = cron
+    {%- endif %}
 {%- if travis_notifications %}
 {{ {"notifications": travis_notifications} | to_yaml }}
 {%- endif %}
--- a/templates/travis/.travis/before_install.sh.j2
+++ b/templates/travis/.travis/before_install.sh.j2
@@ -4,7 +4,11 @@
 
 set -mveuo pipefail
 
-COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+if [[ "$TEST" == "plugin-from-pypi" ]]; then
+  COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
+else
+  COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+fi
 mkdir .travis/vars || true
 echo "---" > .travis/vars/main.yaml
 echo "component_name: {{ plugin_snake }}" >> .travis/vars/main.yaml
--- a/templates/travis/.travis/script.sh.j2
+++ b/templates/travis/.travis/script.sh.j2
@@ -28,6 +28,11 @@ if [ "$TEST" = "docs" ]; then
   exit
 fi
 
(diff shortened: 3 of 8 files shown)


daviddavis:
This is great. Do we need to add an exception for pulpcore?

templates/travis/.travis.yml.j2
```diff
@@ -143,6 +143,10 @@ jobs:
       before_script: skip
       script: bash .travis/cherrypick.sh
     {%- endif %}
+    - stage: test-released-plugin-with-pulpcore-master-branch
+      env:
+        - TEST=plugin-from-pypi
+      if: type = cron
```
fao89, line 149: what happens when we have a breaking change?
Taking drf-yasg to drf-spectacular example, I believe we would break the majority of the CIs with this.
I think we need some conditional for enabling this job

dkliban: @fao89 The goal is to not have breaking changes like that. That's why this check should always be there so plugins can notify pulpcore devs that they made a change that was backwards incompatible.

templates/travis/.travis/script.sh.j2
```diff
@@ -28,6 +28,10 @@ if [ "$TEST" = "docs" ]; then
   exit
 fi
 
+if [[ "$TEST" == "plugin-from-pypi" ]]; then
+  git checkout {{ stable_tag }}
```
mdellweg, line 32: This looks dangerous, as it changes all the files you are already working with in flight.
Maybe you can restrict the checkout to the `tests` directory.

mdellweg (commented):
Tried to remember all points in our discussion.

templates/travis/.travis/before_install.sh.j2
```diff
@@ -5,6 +5,9 @@
 set -mveuo pipefail
 
 COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+if [[ "$TEST" == "plugin-from-pypi" ]]; then
+COMPONENT_VERSION=$(http https://pypi.org/pypi/pulp-file/json | jq -r '.info.version')
```
dkliban, line 9:
```suggestion
COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
```

templates/travis/.travis/script.sh.j2
```diff
@@ -28,6 +28,11 @@ if [ "$TEST" = "docs" ]; then
   exit
 fi
 
+if [[ "$TEST" == "plugin-from-pypi" ]]; then
+  COMPONENT_VERSION=$(http https://pypi.org/pypi/pulp-file/json | jq -r '.info.version')
```
dkliban, line 32:
```suggestion
  COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
```

templates/travis/.travis/before_install.sh.j2
```diff
@@ -4,7 +4,11 @@
 
 set -mveuo pipefail
 
+if [[ "$TEST" == "plugin-from-pypi" ]]; then
+COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
+else
 COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+fi
```
mdellweg, line 11:
```suggestion
if [[ "$TEST" == "plugin-from-pypi" ]]; then
  COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
else
  COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
fi
```

mdellweg (approved):
LGTM

PR #273, opened 2020-09-24
Fix milestone_id issue

fao89:
Release script was breaking when the release did not have changelog entries
[noissue]


quba42 (approved):
I can confirm that applying this to the `pulp_deb` plugin works as expected.
I can also confirm that it fixes the issue of the script failing when there are no issues/changelog entries in the release being created.

PR #275, opened 2020-10-06
Add diagrams to docs/Makefile clean target

quba42:
[noissue]

Also makes full use of the DIAGRAM_SOURCE_DIR variable.

This change was prompted by the following discussion: https://github.com/pulp/pulp_deb/pull/222/files#r500203268


goosemania (approved):
Thanks!

PR #278, opened 2020-10-17
Adding Makefile

daviddavis:
Adding a Makefile for plugin writers to define common build related tasks such as linting, running tests, etc. Inspiration comes from other python projects such as [dynaconf](https://github.com/rochacbruno/dynaconf/blob/master/Makefile), [celery](https://github.com/celery/celery/blob/master/Makefile), and [rq](https://github.com/rq/rq/blob/master/Makefile).

One of the main goals is to allow developers to run the same checks locally as Travis does by using `make`. 

These commands also create a virtual env that isolates the plugin's dev/test dependencies from other plugins as sharing these deps has led to problems. As one example, flake8 plugins automatically run if they are installed but they can differ from plugin to plugin (for example, some plugins are using flake8-docstrings while others don't). 

I'm hoping this will also serve as a place where plugin writers will be able to define more build type tasks related to releasing, etc.

Feedback welcome.

fixes #6938
fixes #5107

### DIFF
--- a/templates/bootstrap/Makefile.j2
+++ b/templates/bootstrap/Makefile.j2
@@ -1,0 +1,50 @@
+COMMIT?=HEAD
+WORKDIR?=.
+VENV?=$(WORKDIR)/.venv
+
+ifeq ($(BIN),)
+ifneq ($(wildcard $(VENV)),)
+BIN=$(VENV)/bin/
+endif
+endif
+
+all: lint test
+
+setup-venv: $(VENV)/bin/activate
+
+$(VENV)/bin/activate:
+	python3 -m venv $(VENV)
+	$(VENV)/bin/pip install -qr dev_requirements.txt
+	$(VENV)/bin/pip install -qr test_requirements.txt
+	$(VENV)/bin/pip install -qe .
+	echo "Set up venv environment at $(VNV)"
+
+lint: lint-black lint-flake8 check-plugin-api check-manifest
+
+lint-black:
+	$(BIN)black --version
+	$(BIN)black --check .
+
+lint-flake8:
+	$(BIN)flake8 --config flake8.cfg
+
+check-commit:
+	python ./.travis/validate_commit_message.py $(COMMIT)
+
+check-plugin-api:
+	./.travis/check_pulpcore_imports.sh
+
+check-manifest:
+	$(BIN)check-manifest
+
+bindings:
+	# TODO: pbindings won't work here. need to recreate this functionality
+	pbindings pulpcore python
+
+test-unit:
+	$(BIN)django-admin test --noinput
+
+test-functional: setup-venv bindings
+	$(BIN)pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional
+
+test: test-unit test-functional
--- a/CHANGES/5107.feature
+++ b/CHANGES/5107.feature
@@ -1,0 +1,1 @@
+Added a Makefile to bootstrap templates to make it easier for plugins to write/run common commands.
--- a/CHANGES/6203.misc
+++ b/CHANGES/6203.misc
@@ -1,0 +1,2 @@
+Adding a ``make test`` command that can be used to run the tests along with all required setup steps
+such as generating bindings.
--- a/CHANGES/6938.feature
+++ b/CHANGES/6938.feature
@@ -1,0 +1,1 @@
+Adding a Makefile to bootstrap template to allow developers to run build-related commands.
--- a/templates/bootstrap/flake8.cfg
+++ b/templates/bootstrap/flake8.cfg
@@ -1,5 +1,5 @@
 [flake8]
-exclude = ./docs/*,*/migrations/*
+exclude = ./docs/*,*/migrations/*,.venv/*
 ignore = W503,Q000,Q003,D100,D104,D106,D200,D202,D205,D400,D401,D402,E203
 max-line-length = 100
 
--- a/templates/bootstrap/pyproject.toml.j2
+++ b/templates/bootstrap/pyproject.toml.j2
@@ -70,6 +70,7 @@ ignore = [
     "HISTORY.rst",
     "dev_requirements.txt",
     "doc_requirements.txt",
+    "Makefile",
     "docs/**",
     "flake8.cfg",
     "template_config.yml",
--- a/templates/travis/.travis/before_install.sh.j2
+++ b/templates/travis/.travis/before_install.sh.j2
@@ -44,24 +44,7 @@ pip install -r dev_requirements.txt
 ./.travis/check_commit.sh
 {%- endif %}
 
-{% if black -%}
-# run black separately from flake8 to get a diff
-black --version
-black --check --diff .
-{%- endif %}
-
-# Lint code.
-flake8 --config flake8.cfg
-
-{% if check_manifest -%}
-# check for any files unintentionally left out of MANIFEST.in
-check-manifest
-{%- endif %}
-
-{% if plugin_name != 'pulpcore' -%}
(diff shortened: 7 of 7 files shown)


templates/bootstrap/Makefile.j2
```diff
@@ -0,0 +1,43 @@
...
+WORKDIR?=.
+VENV?=$(WORKDIR)/.venv
+
+all: lint test
+
+clean:
+	rm -rf $(VENV)
+
+setup-venv:
+	python -m venv $(VENV)
```
mdellweg, line 11: Can we ensure that this will be a python 3 venv?

templates/bootstrap/Makefile.j2
```diff
@@ -0,0 +1,43 @@
+COMMIT?=HEAD
+WORKDIR?=.
+VENV?=$(WORKDIR)/.venv
+
+all: lint test
+
+clean:
+	rm -rf $(VENV)
+
+setup-venv:
```
mdellweg, line 10:
```suggestion
setup-venv: $(VENV)/bin/activate
```

mdellweg:
I like the idea.
I would (eventually) use this Makefile to run the commands in CI, too. (This should help to get consistent behaviour.)

CHANGES/5107.misc
```diff
@@ -0,0 +1 @@
+Adding a Makefile to make it easier to run common plugin commands.
```
ipanova, line 1: s/Adding/Added/

templates/bootstrap/Makefile.j2
mdellweg: Usually, i let the user handle the venv stuff outside of the Makefile.
That helps to reuse the Makefile in CI, where you install different versions of python by other means.

daviddavis: I agree. Putting venv stuff in the Makefile is not optimal. The problem I am trying to solve is that different plugins will have different sets of packages for linting/testing/etc that may conflict. And right now in Pulp dev environments we just use a single virtualenv called 'pulp'. Do you know of a better solution?

mdellweg: Sort of. We can put the parts for linting in another section of the plugin template that can be updated (i.e. not bootstrap).
And reunify them throughout the project.

daviddavis: How do we deal with the situation where a plugins are requiring flake8 plugins that other plugins don’t use? Once a flake8 plugin gets installed in the pulp virtualenv, it runs against all projects.

daviddavis: My one thought is to make `setup-venv` optional and controlled by a variable.

mdellweg: > How do we deal with the situation where a plugins is requiring flake8 plugins that other plugins don’t use? Once a flake8 plugin gets installed in the pulp virtualenv, it runs against all projects.

That is exactly the problem, why running flake in the dev environment is a pain today. I'd like all plugins to require (and install) the same set of flake plugins.

mdellweg: maybe we could add ignore rules that ingore all codes from a plugin with a wildcard?

daviddavis: > That is exactly the problem, why running flake in the dev environment is a pain today. I'd like all plugins to require (and install) the same set of flake plugins.

I feel differently: we should encourage plugins to use the same code styles, etc but not force it upon them. If a plugin wants to use say flake8-mypy they should be free to do so. How would we stop them from doing so?

templates/bootstrap/Makefile.j2
mdellweg: Actually it should be
```suggestion
setup-venv: $(VENV)/bin/activate

$(VENV)/bin/activate:
```

templates/bootstrap/Makefile.j2
mdellweg: I like the "play nice" ansatz. So we should try to solve the root case being that flake plugins are activated only by installing them.

Maybe this helps: https://stackoverflow.com/questions/41860123/how-to-ignore-installed-flake8-plugin-easily-for-one-time

daviddavis: I think having a command to setup a venv in the Makefile and then use it if it exists is a good solution in that it shouldn't affect users who don't want to use a venv. That said I'm not opposed to also having flake8 ignore certain plugins when it's run (or better yet we should probably use `--select`) if we want to also add that.

templates/bootstrap/Makefile.j2
```diff
@@ -0,0 +1,49 @@
...
+all: lint test
+
+setup-venv: $(VENV)/bin/activate
+
+$(VENV)/bin/activate:
+	python3 -m venv $(VENV)
+	$(VENV)/bin/pip install -qr dev_requirements.txt
+	$(VENV)/bin/pip install -qr test_requirements.txt
+	$(VENV)/bin/pip install -qe .
+	echo "Set up venv environment at $(VNV)"
```
mdellweg, line 20:
```suggestion
	echo "Set up venv environment at $(VENV)"
```

dralley:
@daviddavis Should this be closed out, or is it simple enough to get it finished soon?

mdellweg:
I think, this idea should be pursued regardless of it's age.

daviddavis:
I'll try to reduce the scope of this PR (eg not worry about bindings for now) so we can get it merged.

daviddavis:
I'm tempted to dump the venv stuff (at least for now) although this would mean that the problem of having different flake8 plugins for different repos would remain unresolved for now. Thoughts?

dralley:
Totally fine with me

daviddavis:
I'm a bit unsure of how to handle the config variables `black`, `check_manifest`, `check_stray_pulpcore_imports`. I see two options:

1. Remove the config variables and always generate the plugin as if the project requires them. If plugin writers want to remove these, they can do so by updating the Makefile and requirements.txt. This is basically how we handle flake today.
2. Keep them. The downside here is that they are only used to bootstrap the plugin. After that, changing their values in template_config.yml does nothing. This seems misleading.

There might be a third way: create boostrap only options. If that seems like the best approach though, I'd probably go with option 1 here and file a followup issue to add this functionality later.

fao89:
> I'm a bit unsure of how to handle the config variables `black`, `check_manifest`, `check_stray_pulpcore_imports` as the Makefile is bootstrapped and not synced. I see two options:
> 
>     1. Remove the config variables and always generate the plugin as if the project requires them. If plugin writers want to remove these, they can do so by updating the Makefile and requirements.txt. This is basically how we handle flake today.
> 
>     2. Keep them. The downside here is that they are only used to bootstrap the plugin. After that, changing their values in template_config.yml does nothing. This seems misleading.
> 
> 
> There might be a third way: create boostrap only options. If that seems like the best approach though, I'd probably go with option 1 here and file a followup issue to add this functionality later.

I believe all these variables change the CI behavior

daviddavis:
> I believe all these variables change the CI behavior

They won't any longer after this PR. The CI will call `make lint`, etc.

fao89:
Oh! Now I understand!
So I vote for option 1

mdellweg:
Option 3: Move flake config and Makefile into a section that is not bootstrap. They can still land in the same place, but be updated without rebootstrapping.

daviddavis:
Flake configs differ from project to project as plugins are using different sets of flake8 plugins (e.g. some plugins use `flake8-docstrings`). Also, I had imagined that plugins would want to add some of their own custom tasks to Makefile that other plugins might not want (e.g. maybe one plugin wants to use mypy).

daviddavis:
Closing since this is stale. I'd like to move this work forward but

PR #280, opened 2020-10-29
Check version of the plugin under test

mdellweg:
[noissue]

### DIFF
--- a/templates/travis/.travis/start_container.yaml
+++ b/templates/travis/.travis/start_container.yaml
@@ -81,6 +81,14 @@
           command: "docker logs pulp"
           failed_when: true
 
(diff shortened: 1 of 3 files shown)


templates/travis/.travis/start_container.yaml
```diff
@@ -81,6 +81,14 @@
           command: "docker logs pulp"
           failed_when: true
 
+    - name: "Check version of pulgin under test"
```
daviddavis, line 84:
```suggestion
    - name: "Check version of plugin being tested"
```

daviddavis (approved):
This is great. Thank you.

PR #282, opened 2020-11-05
Use constant docker tag name

mdellweg:
CI used to fail if the branch name contained characters not allowed in
container tags.

fixes #7799
https://pulp.plan.io/issues/7799


fao89 (approved):
LGTM!

PR #286, opened 2020-11-18
Match any quotes for the component version check

goosemania:
[noissue]

### DIFF
--- a/templates/travis/.travis/before_install.sh.j2
+++ b/templates/travis/.travis/before_install.sh.j2
@@ -7,7 +7,7 @@ set -mveuo pipefail
 if [[ "$TEST" == "plugin-from-pypi" ]]; then
   COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
 else
-  COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+  COMPONENT_VERSION=$(sed -ne "s/\s*version=['\"]\(.*\)['\"][\s,]*/\1/p" setup.py)
 fi
 mkdir .travis/vars || true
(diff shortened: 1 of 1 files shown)


templates/travis/.travis/before_install.sh.j2
```diff
@@ -7,7 +7,7 @@ set -mveuo pipefail
 if [[ "$TEST" == "plugin-from-pypi" ]]; then
   COMPONENT_VERSION=$(http https://pypi.org/pypi/{{ plugin_dash }}/json | jq -r '.info.version')
 else
-  COMPONENT_VERSION=$(sed -ne 's/\s*version=\"\(.*\)\"[\s,]*/\1/p' setup.py)
+  COMPONENT_VERSION=$(sed -ne "s/\s*version=['\"]\(.*\)['\"][\s,]*/\1/p" setup.py)
```
goosemania, line 10: It's not easy to escape single quotes, so I flipped the outer ones to double ones, because it's easy to escape double quotes. :/

mdellweg: The `"` don't need extra escaping for sed now, because they are part of a `[]` enumeration?

mdellweg (approved):

templates/travis/.travis/before_install.sh.j2
goosemania: The `"` needs escaping, they are escaped. The `'` doesn't need because we are using double quotes outside.
I guess by [definition](http://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Single-Quotes) we can't escape single quotes within single quotes. We can only do concatenation `'\''` to workaround that.

mdellweg: being in single quotes, the \ was not shell escaping, but sed escaping that seems to not be needed anymore because inside [] sed needs a lot less escaping.

PR #292, opened 2020-11-19
Add publish workflow to Github Actions

daviddavis:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -131,8 +131,8 @@ jobs:
       - name: Publish nightly docs
         env:
           BRANCH: ${GITHUB_REF##*/}
+          PULP_DOCS_KEY: {% raw %}${{ secrets.PULP_DOCS_KEY }}{% endraw %}
         run: .ci/scripts/publish_docs.sh nightly $BRANCH
-        shell: bash
       {% endif %}
 
       {%- if test_performance %}
--- a/CHANGES/7880.misc
+++ b/CHANGES/7880.misc
@@ -1,0 +1,1 @@
+Got the publish docs jobs working.
--- a/templates/github/.ci/scripts/publish_docs.sh.j2
+++ b/templates/github/.ci/scripts/publish_docs.sh.j2
@@ -7,8 +7,9 @@ set -euv
 # make sure this script runs at the repo root
 cd "$(dirname "$(realpath -e "$0")")"/../..
 
-openssl aes-256-cbc -K $encrypted_{{ pulpprojectdotorg_key_id }}_key -iv $encrypted_{{ pulpprojectdotorg_key_id }}_iv -in .ci/assets/pulp-infra.enc -out ~/.ssh/pulp-infra -d
-sudo chmod 600 ~/.ssh/pulp-infra
+mkdir ~/.ssh
+echo "$PULP_DOCS_KEY" > ~/.ssh/pulp-infra
+chmod 600 ~/.ssh/pulp-infra
 
 echo "docs.pulpproject.org,8.43.85.236 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBGXG+8vjSQvnAkq33i0XWgpSrbco3rRqNZr0SfVeiqFI7RN/VznwXMioDDhc+hQtgVhd6TYBOrV07IMcKj+FAzg=" >> /home/runner/.ssh/known_hosts
 chmod 644 /home/runner/.ssh/known_hosts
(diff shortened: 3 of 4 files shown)


templates/github/.github/workflows/publish_docs.yml.j2
```diff
@@ -0,0 +1,68 @@
...
+name: Publish docs
+
+on:
+{% if publish_docs_to_pulpprojectdotorg %}
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+{% endif %}
```
daviddavis, line 12: This doesn't work well. It raises an error: `No event triggers defined in 'on'`.

dralley: Is it sufficient to just let this happen nightly?

daviddavis: You mean the workflow or the error?

dralley: The workflow

daviddavis: I don't have a strong opinion here and can change it to only run nightly. Any objections @pulp/continuous-integration?

dralley:
@daviddavis, did you intentially move ```publish_docs.sh``` back to ```.github/workflows/scripts/``` from ```.ci/scripts/```?

daviddavis:
@dralley I did. I meant for `publish_docs.sh` to be specific to Github Actions and serve as the interface to docs-builder.py. wdyt?

dralley:
That's fine, I was just checking since I think it was one of the things that was suggested to be moved to .ci/ previously.

templates/github/.github/workflows/scripts/publish_docs.sh.j2
```diff
@@ -0,0 +1,32 @@
...
+pip3 install -r doc_requirements.txt
+
+export PYTHONUNBUFFERED=1
+export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
+export PULP_SETTINGS=$WORKSPACE/.github/settings/settings.py
+
+eval "$(ssh-agent -s)" #start the ssh agent
+ssh-add ~/.ssh/pulp-infra
+
+if [[ $GITHUB_REF = "refs/heads"* ]]; then
```
dralley, line 26: I'd be more comfortable if this was a script argument rather  than implicit

daviddavis: I think that would make sense if this was a generic script but I meant for this to be specific to the Github Actions environment.

dralley: I know, it just feels hacky.  No strong objection to continuing though

templates/github/.github/workflows/publish_docs.yml.j2
fao89: no objections

dkliban: I find it useful being able to see the built docs right away. What is the concern about the frequency?

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -126,8 +126,10 @@ jobs:
 
       {% if publish_docs_to_pulpprojectdotorg %}
       - name: Publish nightly docs
-        run: .scripts/publish_docs.sh nightly
         shell: bash
+        env:
+          PULP_DOCS_KEY: {% raw %}${{ secrets.PULP_DOCS_KEY }}{% endraw %}
+        run: bash .ci/scripts/publish_docs.sh tag ${GITHUB_REF##*/}
```
fao89, line 132:
```suggestion
        run: bash .ci/scripts/publish_docs.sh nightly ${GITHUB_REF##*/}
```

fao89 (approved):
Thank you!

templates/github/.github/workflows/publish_docs.yml.j2
dralley: My main reasons were

* Releases already trigger a documentation publish which made the "tags" trigger redundant
* If we want to run it any time a branch gets pushed, it doesn't need to be a separate workflow, we can just add it to the normal "ci" workflow and avoid setting up Pulp all over again
* It feels wasteful to have a totally separate job for it, and we should probably try to avoid going crazy with our CI consumption in the future

quba42:
Shouldn't this reference the issue? (https://pulp.plan.io/issues/7880)

templates/github/.github/workflows/publish_docs.yml.j2
daviddavis: I filed an issue here https://pulp.plan.io/issues/7910 to continue the conversation. I'm going to merge this so we can start on the pulp_file 1.4.0 release and then revisit later.

PR #293, opened 2020-11-20
Test plugin bootstrapping

mdellweg:
fixes #8760
https://pulp.plan.io/issues/7860


daviddavis (approved):
Maybe remove .travis.yml.TILT. Other than that, lgtm. :+1:

.github/workflows/bootstrap.yaml
```diff
@@ -0,0 +1,129 @@
...
+        env:
+          - TEST: pulp
+          - TEST: docs
+          - TEST: s3
+
+    # services:
+    #   postgres:
+    #     image: postgres:9.6
+    #   redis:
+    #     image: redis
```
mdellweg, line 71: I don't think we use them. We should remove them from the template too.

PR #296, opened 2020-11-20
Cleanup GitHub template

mdellweg:
Based on #293

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -64,12 +64,6 @@ jobs:
           - TEST: docs
           - TEST: s3
 
-    # services:
-    #   postgres:
-    #     image: postgres:9.6
-    #   redis:
-    #     image: redis
-
     defaults:
       run:
         working-directory: pulp_catdog
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -92,11 +92,11 @@ jobs:
           echo "TEST=${{ matrix.env.TEST }}" >> $GITHUB_ENV
 
       - name: Before Install
-        run: GITHUB_WORKSPACE="${GITHUB_WORKSPACE}/pulp_catdog" .github/workflows/scripts/before_install.sh
+        run: .github/workflows/scripts/before_install.sh
         shell: bash
 
       - name: Install
-        run: GITHUB_WORKSPACE="${GITHUB_WORKSPACE}/pulp_catdog" .github/workflows/scripts/install.sh
+        run: .github/workflows/scripts/install.sh
         env:
           PY_COLORS: '1'
           ANSIBLE_FORCE_COLOR: '1'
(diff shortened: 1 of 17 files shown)


daviddavis (approved):

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -66,12 +66,6 @@ jobs:
           - TEST: docs
           - TEST: s3
 
-    services:
-      postgres:
-        image: postgres:9.6
-      redis:
-        image: redis
```
dralley, line 73: I guess these are "unused" because we build our own container?

daviddavis: Yea, it runs in our single container.

.github/workflows/bootstrap.yaml
```diff
@@ -0,0 +1,123 @@
...
+        # dev_requirements contains tools needed for flake8, etc.
+      - name: Install requirements
+        run: pip3 install -r dev_requirements.txt
+
+      # This check tries to lookup nonexisting commits
+      # - name: Check commit message
+      #   if: github.event_name == 'pull_request'
+      #   env:
+      #     GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
+      #   run: sh .github/workflows/scripts/check_commit.sh
```
fao89, line 36: since we already check commits in another workflow, I think we can remove it

fao89: unless we bring the other workflow to here https://github.com/pulp/plugin_template/blob/master/.github/workflows/check_commits.yaml

mdellweg: Yes, i want to keep it here as a way of documentation.
Whenever we change the workflow template we need to adjust this file too.
And IMHO this comment can save a lot debugging time.

.github/workflows/bootstrap.yaml
```diff
@@ -64,12 +64,6 @@ jobs:
           - TEST: docs
           - TEST: s3
 
-    # services:
-    #   postgres:
-    #     image: postgres:9.6
-    #   redis:
-    #     image: redis
```
daviddavis, line 71: Can this be removed?

daviddavis:
Overall, this LGTM. One small question.

.github/workflows/bootstrap.yaml
mdellweg: Yes, and it is removed... Did you tag the right line?

daviddavis: I meant can you remove the commented out code here?

mdellweg: It is doing exactly that. Removing the comment together with the "real instances" of these services in the workflow templates.

daviddavis: I guess I need a vacation. Sorry.

daviddavis (approved):
This LGTM. @dralley @fao89 did one of you want to also review?

.github/workflows/bootstrap.yaml
```diff
@@ -98,23 +92,23 @@ jobs:
...
           PY_COLORS: '1'
           ANSIBLE_FORCE_COLOR: '1'
         shell: bash
 
       - name: Before Script
         run: |
-          GITHUB_WORKSPACE="${GITHUB_WORKSPACE}/pulp_catdog" .github/workflows/scripts/before_script.sh
-          GITHUB_WORKSPACE="${GITHUB_WORKSPACE}/pulp_catdog" echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.github" >> $GITHUB_ENV
+          .github/workflows/scripts/before_script.sh
+          echo "HTTPIE_CONFIG_DIR=$PWD/.github" >> $GITHUB_ENV
```
fao89, line 108: I think we can remove this since we don't have the config
https://github.com/pulp/pulp_npm/blob/master/.ci/ansible/config.json

when  we used operator on the CI, I don't know why we needed this config for `docs` job not break

fao89 (approved):
LGTM!

.github/workflows/bootstrap.yaml
mdellweg: In some docs builds we try to run the example scripts. I guess it's needed there.

PR #298, opened 2020-11-23
Run on push

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -1,7 +1,13 @@
 {% include 'header.j2' %}
 ---
-name: Pulp PR CI
-on: pull_request
+name: Pulp CI
+on:
+  pull_request:
+    branches:
+      - '*'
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -1,7 +1,13 @@
...
-name: Pulp PR CI
-on: pull_request
+name: Pulp CI
+on:
+  pull_request:
+    branches:
+      - '*'
+  push:
+    branches:
+      - '*'
```
daviddavis, line 10: Does this not work?

```yaml
on: [push, pull_request]
```

fao89: works, but it would run for tags also

daviddavis: Ah I see.

daviddavis (approved):

dralley (approved):

PR #299, opened 2020-11-24
Fix detection of a tag build

mdellweg:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -4,6 +4,20 @@
 
 set -mveuo pipefail
 
+if [ "${GITHUB_REF##refs/heads/}" = "${GITHUB_REF}" ]
+then
+  BRANCH_BUILD=0
+else
+  BRANCH_BUILD=1
+  BRANCH="${GITHUB_REF##refs/heads/}"
+fi
+if [ "${GITHUB_REF##refs/tags/}" = "${GITHUB_REF}" ]
+then
+  TAG_BUILD=0
+else
+  TAG_BUILD=1
+fi
+
 if [[ "$TEST" == "plugin-from-pypi" ]]; then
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -4,6 +4,20 @@
...
+  BRANCH_BUILD=0
+else
+  BRANCH_BUILD=1
+  BRANCH="${GITHUB_REF##refs/heads/}"
+fi
+if [ "${GITHUB_REF##refs/tags/}" = "${GITHUB_REF}" ]
+then
+  TAG_BUILD=0
+else
+  TAG_BUILD=1
```
fao89, line 18: I like it

fao89 (approved):

templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -4,6 +4,20 @@
 
 set -mveuo pipefail
 
+if [ "${GITHUB_REF##refs/heads/}" = "${GITHUB_REF}" ]
```
daviddavis, line 7: I'm curious about what this check is doing. I guess I would have imagined we'd check something like `"$GITHUB_REF" = refs/heads*`?

mdellweg: Well the `##` removes `refs/heads/` from the front if it is there, and in that case the strings stop to be equal. So yes.

daviddavis: Hmm it seems like such an indirect way of checking that, no?

mdellweg: It is compatible with the all bourn shell flavors, afaik.

daviddavis (approved):

PR #303, opened 2020-11-26
Replace pulp image name

fao89:
Tested here:
https://github.com/pulp/pulp_rpm/compare/master...fao89:host
CI:
https://github.com/fao89/pulp_rpm/actions/runs/385900339

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -70,8 +70,8 @@ if [ -n "$PULP_OPENAPI_GENERATOR_PR_NUMBER" ]; then
 fi
 
 cd pulp-openapi-generator
-sed -i -e 's/localhost:24817/pulp/g' generate.sh
-sed -i -e 's/:24817/pulp/g' generate.sh
+sed -i -e 's/localhost:24817/pulp.example.com/g' generate.sh
+sed -i -e 's/:24817/pulp.example.com/g' generate.sh
 cd ..
 
 {% if plugin_name != 'pulpcore' -%}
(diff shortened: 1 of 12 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -70,8 +70,8 @@ if [ -n "$PULP_OPENAPI_GENERATOR_PR_NUMBER" ]; then
 fi
 
 cd pulp-openapi-generator
-sed -i -e 's/localhost:24817/pulp/g' generate.sh
-sed -i -e 's/:24817/pulp/g' generate.sh
+sed -i -e 's/localhost:24817/pulp.example.com/g' generate.sh
+sed -i -e 's/:24817/pulp.example.com/g' generate.sh
```
fao89, line 74: @ipanova you need this at https://github.com/pulp/pulp_container/pull/181

mdellweg:
My impression is, this is helping to hide a problem with a cornercase we found by accident in pulp_container.
I'd rather see us solve the problem there.

fao89:
> My impression is, this is helping to hide a problem with a cornercase we found by accident in pulp_container.
> I'd rather see us solve the problem there.

I agree, but ctrl + F  + new_name is more helpful than ctrl + F  + pulp
image

fao89:
closing in favor of pulp_container approach here https://github.com/pulp/pulp_container/pull/181

ipanova:
Fabricio,  thank you for helping out in finding solution specifically for
pulp_container plugin.  I would still encourage to bring up this image
rename topic to the CI team for the sake of better experience.

PR #311, opened 2020-12-07
Setting secrets dynamically

fao89:
Reference: https://github.com/pulp/pulp_npm/runs/1510991393?check_suite_focus=true#step:5:16

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -66,6 +66,11 @@ jobs:
         run: |
           .github/workflows/scripts/before_script.sh
 
+      - name: Setting secrets
+        run: python3 .github/workflows/scripts/secrets.py "$SECRETS_CONTEXT"
+        env:
+          SECRETS_CONTEXT: {{ "${{ toJson(secrets) }}" }}
+
       - name: Script
         run: .github/workflows/scripts/script.sh
         shell: bash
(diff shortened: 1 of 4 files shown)


templates/github/.ci/scripts/secrets.py.j2
```diff
@@ -0,0 +1,14 @@
...
+secrets = json.loads(sys.argv[1])
+for key, value in secrets.items():
+    print(f"Setting {key} ...")
+    lines = len(value.split("\n"))
+    if lines > 1:
+        os.system(f"echo '{key}<<EOF' >> $GITHUB_ENV")
+        os.system(f"echo '{value}' >> $GITHUB_ENV")
+        os.system("echo 'EOF' >> $GITHUB_ENV")
+    else:
+        os.system(f"echo '{key}={value}' >> $GITHUB_ENV")
```
daviddavis, line 14: I would probably move this file to templates/github/.github/scripts/secrets.py.j2 since it's specific to github.

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -66,6 +66,12 @@ jobs:
         run: |
           .github/workflows/scripts/before_script.sh
 
+      - name: Setting secrets
+        if: github.event_name != 'pull_request'
```
daviddavis, line 70: You can probably remove this line.

templates/github/.github/workflows/release.yml.j2
```diff
@@ -62,6 +62,12 @@ jobs:
         run: |
           .github/workflows/scripts/before_script.sh
 
+      - name: Setting secrets
+        if: github.event_name != 'pull_request'
```
daviddavis, line 66: This line too.

templates/github/.github/workflows/nightly.yml.j2
fao89: I believe removing it would break the CI, something similar broke pulp_ansible: https://github.com/pulp/pulp_ansible/runs/1444669298?check_suite_focus=true#step:8:9

daviddavis: The build you link to is not a nightly build. I think that the code here in `nightly.yml` should not affect non-nightly builds?

daviddavis: Also I don't think `github.event_name` would ever be `pull_request` here.

fao89: true, it was a bad ctrl+c from ci.yml

daviddavis (approved):

PR #312, opened 2020-12-07
Added a deprecation check for moved or removed files

daviddavis:


fao89 (approved):
sweet!

plugin-template
```diff
@@ -57,6 +58,13 @@ DEFAULT_SETTINGS = {
 }
 
 
+DEPRECATED_FILES = [
+    ".ci/scripts/release_requirements.txt",
+    # ".travis",
+    # ".travis.yml",
+]
```
mdellweg, line 65: Do you think, we can call out any file we didn't touch, instead of keeping an ever growing list of files we "know" about?

"pre_/post_" scripts are a different thing though.
But files needed by certain flags in the plugin template config could be reported too.

daviddavis: It would be nice to not keep a list but I am not sure how this would work. One thing that comes to mind is we could just check certain directories but I was hoping to deprecate `.travis` and `.travis.yml` for example.

Also, what if a plugin adds a file to `.ci`? Then they would get a warning every time they run the plugin?

mdellweg: So they'd need to be able to provide another list of "regular" files.
Seems like a mess.
We can think about this later, and i am not going to block this PR.

mdellweg (approved):

PR #320, opened 2020-12-08
Remove constants.py file

daviddavis:
I can't figure out what this file is used for?


fao89:
I didn't know this file exist

mdellweg (approved):
Tests pass, so we should be confident this file is not needed to bootstrap any section.

PR #323, opened 2020-12-14
Run performance tests

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -168,8 +168,15 @@ jobs:
     runs-on: ubuntu-latest
     needs: test
 
-    env:
-      TEST: performance
+    {%- if test_performance is iterable and test_performance|length > 1 %}
+    strategy:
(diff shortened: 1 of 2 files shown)


dkliban (commented):
so what happens when test_performance is not an iterable?

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -182,16 +189,16 @@ jobs:
...
-        with:
-          ruby-version: "2.6"
-
       - name: Install httpie
         run: |
           echo ::group::HTTPIE
           sudo apt-get update -yq
           sudo -E apt-get -yq --no-install-suggests --no-install-recommends install httpie
           echo ::endgroup::
+          echo "TEST=performance" >> $GITHUB_ENV
```
fao89, line 198: @dkliban when it is not iterable, this line is everything it needs, here you can see the test: https://github.com/fao89/pulp_file/runs/1557393762?check_suite_focus=true

dkliban (approved):

PR #324, opened 2020-12-14
Don't publish docs to the root if not the latest version

dkliban:
This patch updates the script that publishes docs to pulpproject.org. The version of the docs being
published is compared to the version of the package on PyPI. If the latest version of the package on
PyPI is equal to the version of the docs being published, the site root is updated with the docs.
Otherwise the docs are published to 3.y and 3.y.z directories.

re: #7766
https://pulp.plan.io/issues/7766

### DIFF
--- a/templates/github/.ci/scripts/docs-builder.py.j2
+++ b/templates/github/.ci/scripts/docs-builder.py.j2
@@ -8,6 +8,9 @@ import os
 import re
 from shutil import rmtree
 import tempfile
+import requests
+import json
+from packaging import version
 
 WORKING_DIR = os.environ["WORKSPACE"]
 
@@ -85,8 +88,17 @@ def main():
 
     ga_build = False
 
+    publish_at_root = False
+
     if (not re.search("[a-zA-Z]", branch) or "post" in branch) and len(branch.split(".")) > 2:
         ga_build = True
+        # Only publish docs at the root if this is the latest version
+        r = requests.get("https://pypi.org/pypi/{{ plugin_dash }}/json")
+        latest_version = version.parse(json.loads(r.text)["info"]["version"])
+        docs_version = version.parse(branch)
+        if latest_version == docs_version:
+            publish_at_root = True
+        # Post releases should use the x.y.z part of the version string to form a path
         if "post" in branch:
             branch = ".".join(branch.split(".")[:-1])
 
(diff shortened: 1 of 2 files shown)


templates/github/.ci/scripts/docs-builder.py.j2
```diff
@@ -85,8 +88,17 @@ def main():
 
     ga_build = False
 
+    publish_at_root = False
+
     if (not re.search("[a-zA-Z]", branch) or "post" in branch) and len(branch.split(".")) > 2:
         ga_build = True
+        # Only publish docs at the root if this is the latest version
+        r = requests.get("https://pypi.org/pypi/{{ plugin_dash }}/json")
+        latest_version = version.parse(json.loads(r.text)["info"]["version"])
```
fao89, line 97: cool!

fao89 (approved):

templates/github/.ci/scripts/docs-builder.py.j2
```diff
@@ -8,6 +8,9 @@ import os
 import re
 from shutil import rmtree
 import tempfile
+import requests
+import json
+from packaging import version
```
mdellweg, line 13: :+1: Yup, that's the right one...

templates/github/.ci/scripts/docs-builder.py.j2
```diff
@@ -85,8 +88,17 @@ def main():
...
     ga_build = False
 
+    publish_at_root = False
+
     if (not re.search("[a-zA-Z]", branch) or "post" in branch) and len(branch.split(".")) > 2:
         ga_build = True
+        # Only publish docs at the root if this is the latest version
+        r = requests.get("https://pypi.org/pypi/{{ plugin_dash }}/json")
+        latest_version = version.parse(json.loads(r.text)["info"]["version"])
+        docs_version =  version.parse(branch)
```
mdellweg, line 98:
```suggestion
        docs_version = version.parse(branch)
```
It's correct that this branch-version still contains the post section, because it needs to compare to the package version on pypi, right?

mdellweg (approved):

templates/github/.ci/scripts/docs-builder.py.j2
dkliban: Yes, exactly.

PR #328, opened 2021-01-05
Fix commit message issue

fao89:
issue: https://github.com/pulp/plugin_template/actions/runs/463579136

### DIFF
--- a/.ci/bootstrap_catdog.sh
+++ b/.ci/bootstrap_catdog.sh
@@ -2,7 +2,13 @@
 
 set -euv
 
-COMMIT_MSG=$(git log --format=%B -n 1 HEAD^2)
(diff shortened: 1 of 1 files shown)


mdellweg:
What is this needed for?
Can we make the script not depend on GH-Actions?

fao89:
> What is this needed for?
> Can we make the script not depend on GH-Actions?

passing through "Required PR" commit messages, and I think it is possible to make it not GHA dependent

mdellweg (approved):
If i understand correctly, this is to prevent using the required-PR mechanism for anything but pull requests.

fao89:
> If i understand correctly, this is to prevent using the required-PR mechanism for anything but pull requests.

yes, that was the idea!

PR #341, opened 2021-02-17
Change the version check to use app label

mdellweg:
A change in pulpcore (status API) requires us to change the version test
to use the app label instead of the repository name.
Along with this change we introduce a legacy lookup as a fallback to not
break all CI at the same time.

[noissue]

### DIFF
--- a/templates/github/.ci/ansible/start_container.yaml
+++ b/templates/github/.ci/ansible/start_container.yaml
@@ -81,13 +81,22 @@
           command: "docker logs pulp"
           failed_when: true
 
-    - name: "Check version of component being tested"
-      assert:
-        that:
-          - (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] | canonical_semver == (component_version | canonical_semver)
-        fail_msg: |
-          Component {{ component_name }} was expected to be installed in version {{ component_version }}.
-          Instead it is reported as version {{ (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] }}.
+    - block:
+        - name: "Check version of component being tested"
+          assert:
+            that:
+              - (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] | canonical_semver == (component_version | canonical_semver)
+            fail_msg: |
+              Component {{ component_name }} was expected to be installed in version {{ component_version }}.
+              Instead it is reported as version {{ (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] }}.
+      rescue:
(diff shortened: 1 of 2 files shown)


bmbouter:
Maybe I applied it wrong, but I think the failed when I tried to use it with pulpcore. See the failures here: https://github.com/pulp/pulpcore/pull/1102

templates/github/.ci/ansible/start_container.yaml
```diff
@@ -81,13 +81,26 @@
...
-        that:
-          - (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] | canonical_semver == (component_version | canonical_semver)
-        fail_msg: |
-          Component {{ component_name }} was expected to be installed in version {{ component_version }}.
-          Instead it is reported as version {{ (result.json.versions | items2dict(key_name="component", value_name="version"))[component_name] }}.
+    - block:
+        - name: "Check the component is in status"
+          assert:
+            that:
+              - component_name in (result.json.versions | items2dict(key_name="component", value_name="version") | keys)
```
mdellweg, line 88: So the keys filter does not exist... still the rescue block is executed.
Then we should be fine just removing this task.

mdellweg: I was so laser focused seeing a failure, that i saw a failure, not realizing it was the wrong one.

mdellweg:
@bmbouter Sorry, this version should work better. Please try again.

And thank you for not blindly trusting it, but insisting to apply it to your PR first.

bmbouter:
I think it worked, but this time it's failing not on finding pulpcore, but on pulp_file (in the pulpcore CI). Any advice on what I should change to have that set the vars correctly too? I didn't expect it to check pulp_file also.

https://github.com/pulp/pulpcore/pull/1102/checks?check_run_id=1926773090

mdellweg:
> I think it worked, but this time it's failing not on finding pulpcore, but on pulp_file (in the pulpcore CI). Any advice on what I should change to have that set the vars correctly too? I didn't expect it to check pulp_file also.
> 
> https://github.com/pulp/pulpcore/pull/1102/checks?check_run_id=1926773090

That is not the check you fixed. But in the meantime the CLI tests were added to this CI. And the cli still uses the old names.

bmbouter:
Ah yes! Let's collab at our time tomorrow and figure out how to fix. Ty!

bmbouter (approved):
I successfully used this PR in my pulpcore PR, so I'm +1 to merge.

PR #342, opened 2021-02-19
List pip dependencies

fao89:
[noissues]

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -8,6 +8,8 @@ REPO_ROOT="$PWD"
 
 set -euv
 
+source .github/workflows/scripts/utils.sh
+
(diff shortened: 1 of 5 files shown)


daviddavis:
I think you need to add `source .github/workflows/scripts/utils.sh` to install.sh.j2.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -143,3 +143,8 @@ fi
 
 ansible-playbook build_container.yaml
 ansible-playbook start_container.yaml
+
+echo ::group::PIP_LIST
+source .github/workflows/scripts/utils.sh
```
daviddavis, line 148: This needs to go before `cd .ci/ansible/`.

daviddavis:
I applied this to pulpcore: https://github.com/pulp/pulpcore/pull/1148

daviddavis (approved):
Looks like this is working really well ^^^

:1st_place_medal:

PR #345, opened 2021-02-24
Adding nightly FIPS tests

fao89:
https://pulp.plan.io/issues/3800
closes #3800

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -48,6 +48,7 @@ DEFAULT_SETTINGS = {
     'redmine_project': None,
     'test_bindings': True,
     'test_cli': False,
+    'test_fips_nightly': False,
     'test_performance': False,
     'test_released_plugin_with_next_pulpcore_release': False,
     'test_s3': False,
--- a/templates/github/.github/workflows/fips.yml.j2
+++ b/templates/github/.github/workflows/fips.yml.j2
@@ -1,0 +1,70 @@
+{% include 'header.j2' %}
+---
+name: Pulp FIPS CI
+on:
+  {% if test_fips_nightly -%}
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # runs at 4:00 UTC daily
+    - cron: '00 4 * * *'
+  {%- else -%}
+  workflow_dispatch:
+  {%- endif %}
+
+jobs:
+  vagrant:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test_type: pulp3-source-centos8-fips
+          - test_type: pulp3-source-centos7-fips
+    steps:
+      - uses: actions/checkout@v2.3.1
+        with:
+          submodules: true
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Ansible
+        run: |
+          pip3 install -U pip
+          pip3 install "ansible>2.9,<3"
+          ansible --version
+      - name: Pulp Installer
+        run: |
+          cd ..
+          git clone --depth=1 https://github.com/pulp/pulp_installer
+      - name: Install Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer/
+          git submodule update --init
+          .github/workflows/scripts/install.sh {{ "${{ matrix.test_type }}" }}
+          {% if plugin_name != 'pulpcore' -%}
+          git clone https://github.com/pulp/pulpcore
+          {%- else -%}
+          git clone https://github.com/pulp/pulp_file
+          {%- endif %}
+      - name: Run Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer
+          {% if plugin_name not in ['pulp_file', 'pulpcore'] -%}
+          sed -e "s/pulp-file/{{ plugin_dash }}/g" -e "s/pulp_file/{{ plugin_name }}/g" -e "/pulp_source_dir/a pulp_webserver_disable_https: true" example.dev-config.yml > local.dev-config.yml
+          {%- else -%}
+          sed -e "/pulp_source_dir/a pulp_webserver_disable_https: true" example.dev-config.yml > local.dev-config.yml
+          {%- endif %}
+          echo "vagrant ssh \$1 -c \"workon pulp && cd devel/{{ plugin_name }} && pip install -r functest_requirements.txt\"" >> .github/workflows/scripts/script.sh
+          echo "vagrant ssh \$1 -c \"workon pulp && cd devel/{{ plugin_name }}/ && pytest -v -r sx --color=yes --pyargs {{ plugin_name }}.tests.functional\"" >> .github/workflows/scripts/script.sh
+          .github/workflows/scripts/script.sh {{ "${{ matrix.test_type }}" }}
+        env:
+          PY_COLORS: '1'
+          ANSIBLE_FORCE_COLOR: '1'
+      - name: After failure
+        if: failure()
+        run: |
+          cd ../pulp_installer
+          .github/workflows/scripts/after_failure.sh {{ "${{ matrix.test_type }}" }}
--- a/CHANGES/3800.feature
+++ b/CHANGES/3800.feature
@@ -1,0 +1,1 @@
+Adding nightly FIPS tests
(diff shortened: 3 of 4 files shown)


fao89:
I'll be testing on pulp_npm:
https://github.com/pulp/pulp_npm/pull/72

fao89:
could see it running here:
https://github.com/fao89/pulp_npm/runs/1981133370?check_suite_focus=true

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,64 @@
+{% include 'header.j2' %}
+---
+name: Pulp FIPS Nightly CI
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # runs at 4:00 UTC daily
+    - cron: '00 4 * * *'
```
daviddavis, line 8: I imagine that not all plugins will want to test against FIPS nightly. What if we introduce a setting (default to true?) and then do something like this:

```suggestion
  {% if fips_test %}
  schedule:
    # * is a special character in YAML so you have to quote this string
    # runs at 4:00 UTC daily
    - cron: '00 4 * * *'
  {% else %}
  workflow_dispatch:
  {% endif %}
```

daviddavis:
This work looks great. :100: 

I see that the centos 7 test is failing against pulp_npm and pulp_ansible?

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,64 @@
...
+          sed -i "/{{ plugin_name }}/d" .github/workflows/scripts/install.sh
+          {%- endif %}
+          .github/workflows/scripts/install.sh {{ "${{ matrix.test_type }}" }}
+      - name: Run Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer
+          {%- if plugin_name not in ['pulp_file', 'pulpcore'] %}
+          sed -i "s/pulp-file/{{ plugin_dash }}/g" example.dev-config.yml
+          sed -i "s/pulp_file/{{ plugin_name }}/g" example.dev-config.yml
```
mdellweg, line 51:
```suggestion
          sed -e "s/pulp-file/{{ plugin_dash }}/g" -e "s/pulp_file/{{ plugin_name }}/g" example.dev-config.yml > local.dev-config.yml
```

mdellweg: You can perform multiple scripts in one `sed` call. Also i prefer not to change git controlled files (but that is a matter of taste).

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          cd ..
+          git clone --depth=1 https://github.com/pulp/pulp_installer
+      - name: Install Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer/
+          git submodule update --init
+          {%- if plugin_name in ['pulp_file', 'pulpcore'] %}
+          sed -i "/{{ plugin_name }}/d" .github/workflows/scripts/install.sh
```
mdellweg, line 42: This looks dangerous to me. Looking at these lines here, it's not even clear to me, what lines are supposed to be deleted. And if that file introduces a new important line containing "pulpcore" it will break in a way that is almost impossible to debug.

templates/github/.github/workflows/fips.yml.j2
fao89: sweet!

fao89:
> This work looks great. 
> 
> I see that the centos 7 test is failing against pulp_npm and pulp_ansible?

yes, it failed all over the weekend, I've noticed GHA on my fork is behaving differently,
on my fork, it is already on ubuntu 20, so I pinned ubuntu 20 here expecting to get the same results, but it didn't work.
I suspect my fork is using a most updated runner

fao89:
Actually, they are using the same version

[pulp](https://github.com/pulp/pulp_npm/runs/2001123779?check_suite_focus=true):
image

[fork](https://github.com/fao89/pulp_npm/runs/1981133370?check_suite_focus=true):
image

I need to investigate more

templates/github/.github/workflows/fips.yml.j2
fao89: I agree, it currently deletes git clone of pulpcore and pulp_file, I plan to change the script on the installer side, make it more customizable there

templates/github/.github/workflows/fips.yml.j2
fao89: Nice! I didn't know about that

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,67 @@
...
+          {%- endif %}
+          .github/workflows/scripts/install.sh {{ "${{ matrix.test_type }}" }}
+      - name: Run Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer
+          {%- if plugin_name not in ['pulp_file', 'pulpcore'] %}
+          sed -e "s/pulp-file/{{ plugin_dash }}/g" -e "s/pulp_file/{{ plugin_name }}/g" example.dev-config.yml > local.dev-config.yml
+          {%- endif %}
+          sed -e "/pulp_source_dir/a pulp_webserver_disable_https: true" example.dev-config.yml > local.dev-config.yml
```
mdellweg, line 56: Sorry i didn't see this in all it's beauty. This second sed will overwrite the results of the first one.
I guess you'll need to add it to the first and wrap that second in an else.

daviddavis:
LGTM. @mdellweg did you want to review again before we merge?

plugin-template
```diff
@@ -48,6 +48,7 @@ DEFAULT_SETTINGS = {
     'redmine_project': None,
     'test_bindings': True,
     'test_cli': False,
+    'fips_test': True,
```
mdellweg, line 51: Is True the proper default here? I know we want that on most plugins, but usually we add features disabled by default here.
(e.g. `redmine_project`, `test_cli`, `test_performance`, `test_released_plugin_with_next_pulpcore_release`, `test_s3`, ...)

edit: I see, that i :+1:d the idea to add it True... So don't see this as a blocker.

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,68 @@
...
+---
+name: Pulp FIPS Nightly CI
+on:
+  {% if fips_test -%}
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # runs at 4:00 UTC daily
+    - cron: '00 4 * * *'
+  {%- else -%}
+  workflow_dispatch:
```
mdellweg, line 11: Does this mean, that you can still trigger the test by hand?
Can we rename the variable to `test_fips_nightly` to reflect that?

templates/github/.github/workflows/fips.yml.j2
mdellweg: Being customizable in the script is great, because then you can see there that there might be an expected interface to it.

templates/github/.github/workflows/fips.yml.j2
```diff
@@ -0,0 +1,68 @@
...
+          {%- if plugin_name in ['pulp_file', 'pulpcore'] %}
+          sed -i "/{{ plugin_name }}/d" .github/workflows/scripts/install.sh
+          {%- endif %}
+          .github/workflows/scripts/install.sh {{ "${{ matrix.test_type }}" }}
+      - name: Run Vagrant
+        shell: bash
+        run: |
+          cd ../pulp_installer
+          {% if plugin_name not in ['pulp_file', 'pulpcore'] -%}
+          sed -e "s/pulp-file/{{ plugin_dash }}/g" -e "s/pulp_file/{{ plugin_name }}/g" -e "/pulp_source_dir/a pulp_webserver_disable_https: true" example.dev-config.yml > local.dev-config.yml
```
mdellweg, line 54: :+1:
Maybe in a next iteration, we should just provide the proper dev-config in the assets directory. It can use the additional_plugins variable at templating to determine what to install.

mdellweg (approved):
Rereview looks good.

plugin-template
daviddavis: I suggested true by default but i think you’re right.

templates/github/.github/workflows/fips.yml.j2
fao89: I've started a PR: https://github.com/pulp/pulp_installer/pull/539

mdellweg (approved):

PR #346, opened 2021-03-01
Fix publish jobs by using app label

daviddavis:
fixes #8311

### DIFF
--- a/CHANGES/8311.bugfix
+++ b/CHANGES/8311.bugfix
@@ -1,0 +1,1 @@
+Fix publish job by using app label.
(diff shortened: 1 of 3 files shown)


fao89 (approved):
👏🏽

mdellweg (approved):
This looks right to me.
But i keep wondering, whether that change will imply that the client packages are named differently, or if there's some other unforseen side effect?

daviddavis:
> But i keep wondering, whether that change will imply that the client packages are named differently, or if there's some other unforseen side effect?

It's possible but at least from my reading of the code, it looks like we'll continue to use the package name. For one, it looks like we use [the module name to filter the api spec](https://github.com/pulp/pulpcore/blob/28b875bcf2d24ff23c64275844d817dfb3d1a9dc/pulpcore/openapi/__init__.py#L373).

daviddavis:
Oh but maybe module name != plugin name. That could be problematic?

mdellweg:
Hmm. I may be OK. Let's try for one day.

bmbouter:
Thank you for fixing this.

PR #347, opened 2021-03-01
Prevent collision with PULP_CONTAINER env variable

daviddavis:
See example of the collision: https://git.io/Jtpwj

[noissue]


goosemania (approved):
Thanks, it works.

fao89 (approved):

PR #348, opened 2021-03-02
Add check for mixed usage of gettext and f-strings

daviddavis:
fixes #8316


mdellweg (approved):
Nice one!
Isn't this the kind of thing flake8 should do for us?

daviddavis:
> Isn't this the kind of thing flake8 should do for us?

Perhaps I suppose?

PR #354, opened 2021-03-11
Added issue_tracker config

gerrod3:
[noissue]

Adding new config option, plus a --migrate-github-issues command to help plugins move off of Redmine in the future. This PR follows the instructions @daviddavis came up with here: https://hackmd.io/y3AnY_bCQCODemWQk2bfoQ?both. I'm going to test this out soon on Pulp Python after its next release.

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -30,6 +30,7 @@ DEFAULT_SETTINGS = {
     'deploy_to_pypi': True,
     'docker_fixtures': False,
     'docs_test': True,
+    'issue_tracker': 'redmine',
     'plugin_app_label': None,
     'plugin_camel': None,
     'plugin_camel_short': None,
@@ -117,6 +118,10 @@ def main():
                             Include more output.
 
                         '''))
+    parser.add_argument('--migrate-github-issues', action='store_true',
+                        help=textwrap.dedent('''\
+                            Migrate plugin's issue checker CI/CD from Redmine to GitHub
+                        '''))
     args = parser.parse_args()
 
 
@@ -164,6 +169,13 @@ def main():
     # Config key is used by the template_config.yml.j2 template to dump
     # the config. (note: uses .copy() to avoid a self reference)
     config['config'] = config.copy()
+
+    if args.migrate_github_issues:
+        if config['issue_tracker'] == 'github':
+            print(f"{args.plugin_name} is already using GitHub as the issue tracker")
+            return 2
+        migrate_github_issues(config, plugin_root_dir, verbose=args.verbose)
+        return 0
 
     if args.github or args.all:
         if not config['pypi_username'] and (config['deploy_client_to_pypi'] or \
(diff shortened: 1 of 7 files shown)


templates/bootstrap/pyproject.toml.j2
```diff
@@ -11,7 +11,11 @@ filename = "CHANGES.rst"
 directory = "CHANGES/"
 title_format = "{version} ({project_date})"
 template = "CHANGES/.TEMPLATE.rst"
+{% if issue_tracker=='redmine' %}
 issue_format = "`#{issue} <https://pulp.plan.io/issues/{issue}>`_"
+{% else %}
+issue_format = "`#{issue} <https://github.com/pulp/{{ plugin_name }}/issues/{issue}>`_"
```
mdellweg, line 17: Should we add support for different orgs here?

templates/bootstrap/README.md.j2
```diff
@@ -8,10 +8,15 @@ For more information, please see the [documentation](docs/index.rst) or the [Pul
 How to File an Issue
 --------------------
 
+{% if issue_tracker=='redmine' %}
 Visit [our tracker](https://pulp.plan.io/projects/{{ redmine_project }}/issues/new) to file a {{ plugin_dash }} issue.
+{% else %}
+File through this project's GitHub issues and appropriate labels.
+{% endif %}
```
mdellweg, line 15: Maybe we should allow not to specify any issue tracker too.

plugin-template
```diff
@@ -307,5 +319,23 @@ def check_for_deprecated_files(plugin_root_dir):
         print("No deprecated files found.")
 
 
+def migrate_github_issues(config, plugin_root_dir, verbose=False):
+    """Moves the plugin's issue checker CI/CD off of Redmine and onto GitHub"""
+    config['issue_tracker'] = 'github'
+    name = config['plugin_name']
+    redmine = config['redmine_project'] or name
```
fao89, line 326: I think it should explicitly raise an error instead of using `or`

templates/bootstrap/pyproject.toml.j2
fao89: I think so

plugin-template
```diff
@@ -307,5 +319,23 @@ def check_for_deprecated_files(plugin_root_dir):
...
+    write_template_section(config, "generate_config", plugin_root_dir, verbose=verbose)
+    write_template_section(config, "github", plugin_root_dir, verbose=verbose)
+    print("\nUpdating pyproject.toml to point to GitHub issues")
+    search_string = "pulp.plan.io"
+    replace_string = f"github.com\/pulp\/{config['plugin_name']}"
+    os.system(f"sed -i 's/{search_string}/{replace_string}/' {plugin_root_dir}/pyproject.toml")
+    print(f"\nPlease check the Pulp-CI PR processor (https://git.io/Jtsz5) and remove {name}"
+          f" if present")
+    print(f"\nDon't forget to migrate current issues from https://pulp.plan.io/projects/{redmine}"
+          f" and remove the repo from the project settings")
```
fao89, line 338: do we have some script for it?
If so, can we link it here?

templates/bootstrap/pyproject.toml.j2
gerrod3: I think this might require a separate PR. There are other templates where we have Github links with the Pulp org.

plugin-template
gerrod3: This is the script I found https://github.com/IQSS/redmine2github that I was talking about in our previous meeting. I haven't tested it, but would you still like me to add a link to it?

plugin-template
```diff
@@ -307,5 +319,27 @@ def check_for_deprecated_files(plugin_root_dir):
...
+    redmine = config['redmine_project']
+    if not redmine:
+        print("Redmine project name not set. Please set 'redmine_project' in template_config.yml"
+              " and rerun this command")
+        return 2
+    config['redmine_project'] = None
+    config['update_redmine'] = False
+    write_template_section(config, "generate_config", plugin_root_dir, verbose=verbose)
+    write_template_section(config, "github", plugin_root_dir, verbose=verbose)
+    print("\nUpdating pyproject.toml to point to GitHub issues")
```
mdellweg, line 335: Why should a print statement start with a newline?

plugin-template
```diff
@@ -307,5 +319,27 @@ def check_for_deprecated_files(plugin_root_dir):
         print("No deprecated files found.")
 
 
+def migrate_github_issues(config, plugin_root_dir, verbose=False):
+    """Moves the plugin's issue checker CI/CD off of Redmine and onto GitHub"""
+    config['issue_tracker'] = 'github'
```
mdellweg, line 324:
```suggestion
    print("Updating template_config.yml and and templating ci section")
    config['issue_tracker'] = 'github'
```

mdellweg (commented):
So should we merge this before or after it has been tested with pulp_ansible?

plugin-template
gerrod3: It was done earlier in the script. I changed it to after.

gerrod3:
> So should we merge this before or after it has been tested with pulp_ansible?

I already testing this with pulp_python https://github.com/pulp/pulp_python/pull/355

mdellweg (approved):

plugin-template
fao89: sorry, I forgot to reply, if it is not tested, it is better not to adding it

fao89 (approved):
LGTM!

PR #356, opened 2021-03-15
Partially fix docs publishing job

fao89:
It fixes:
```
UnboundLocalError: local variable 'x_y_version' referenced before assignment
```
but the real issue is checking if we are publishing the latest version:
https://github.com/pulp/plugin_template/blob/master/templates/github/.ci/scripts/docs-builder.py.j2#L94-L100

### DIFF
--- a/templates/github/.ci/scripts/docs-builder.py.j2
+++ b/templates/github/.ci/scripts/docs-builder.py.j2
@@ -96,7 +96,10 @@ def main():
         r = requests.get("https://pypi.org/pypi/{{ plugin_dash }}/json")
         latest_version = version.parse(json.loads(r.text)["info"]["version"])
         docs_version = version.parse(branch)
-        if latest_version == docs_version:
+        # This is to mitigate delays on PyPI which doesn't update metadata in timely manner.
+        # It doesn't prevent incorrect docs being published at root if 2 releases are done close
+        # to each other, within PyPI delay. E.g. Release 3.11.0 an then 3.10.1 immediately after.
+        if docs_version >= latest_version:
             publish_at_root = True
         # Post releases should use the x.y.z part of the version string to form a path
         if "post" in branch:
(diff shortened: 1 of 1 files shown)


goosemania:
@fao89 , wait for @ipanova , she has an idea to improve the check, it won't solve all the corner cases but the majority

ipanova:
The problem we are facing happens because pypi does not have updated metadata yet and so this condition fails, because `latest_version` is still not updated.
https://github.com/pulp/plugin_template/pull/356/files#diff-ae034e553d7c235ba1ec1e1dbf724ada77eb55a672cff2c5a94970482a2926a1R99
We can improve the logic this way:
```
latest_version = version.parse(json.loads(r.text)["info"]["version"])
docs_version = version.parse(branch)
if  docs_version >= latest_version:
        publish_at_root = True
```
This logic still would not cover the case when, for example we have 3.10.0 as current pulpcore version. We have just released 3.11.0 and right away started the release of 3.10.1. As a result 3.10.1 would override the 3.11.0 docs. To avoid this we should not have 2 consecutive releases maybe wait some couple of hours in between.

mdellweg:
Is it possible to not depend on pypi at all for this check? We have rsync access to the docs location, so if we could gather information about the version documented in that root directory, we could use that, no?

fao89:
> Is it possible to not depend on pypi at all for this check? We have rsync access to the docs location, so if we could gather information about the version documented in that root directory, we could use that, no?

+1, for me this is the best way to go

mdellweg:
> > Is it possible to not depend on pypi at all for this check? We have rsync access to the docs location, so if we could gather information about the version documented in that root directory, we could use that, no?
> 
> +1, for me this is the best way to go

I'm not saying we should not merge this, if it improves the situation today.

fao89:
This PR "fixes" the problem,
the CI won't fail, but when you go to https://docs.pulpproject.org it won't display the latest doc
you would need to go to https://docs.pulpproject.org/pulpcore/en/3.11.0/

So basically it fixes the CI, not the problem

goosemania:
I suggest to incorporate Ina's suggestion and then merge all together.
I think it will solve 99% of the problems for now, it will also display latest docs.

Everyone is onboard with not relying on PyPI for a check, that's clear. The question is how.

fao89:
> I suggest to incorporate Ina's suggestion (`docs_version >= latest_version`) and then merge all together.
> I think it will solve 99% of the problems for now, it will also display latest docs.
> 
> Everyone is onboard with not relying on PyPI for a check, that's clear. The question is how.

I just incorporate it

templates/github/.ci/scripts/docs-builder.py.j2
```diff
@@ -96,7 +96,7 @@ def main():
         r = requests.get("https://pypi.org/pypi/{{ plugin_dash }}/json")
         latest_version = version.parse(json.loads(r.text)["info"]["version"])
         docs_version = version.parse(branch)
-        if latest_version == docs_version:
+        if docs_version >= latest_version:
```
goosemania, line 99: Could you add comment please - `This is to mitigate delays on PyPI which doesn't update metadata in timely manner. It doesn't prevent incorrect docs being published at root if 2 releases are done close to each other, within PyPI delay. E.g. Release 3.11.0 an then 3.10.1 immediately after.`
I'm worried that we'll forget all those details.

goosemania (approved):

ipanova (approved):
looks good!

PR #357, opened 2021-03-18
Create a manual trigger for the release process

daviddavis:
fixes #8404


templates/github/.ci/scripts/publish_plugin_pypi.sh.j2
```diff
@@ -7,6 +7,14 @@ cd "$(dirname "$(realpath -e "$0")")"/../..
 
 set -euv
 
+export VERSION=$(http pulp/pulp/api/v3/status/ | jq --arg plugin {{ plugin_app_label }} -r '.versions[] | select(.component == $plugin) | .version')
+export response=$(curl --write-out %{http_code} --silent --output /dev/null https://pypi.org/project/pulp-npm-client/$VERSION/)
+if [ "$response" == "200" ];
+then
+  echo "{{ plugin_name }} $VERSION has already been released. Skipping."
+  exit
+fi
```
daviddavis, line 16: This code is basically copied over from the other release scripts such as publish_client_pypi.sh.

daviddavis:
Example run: https://github.com/pulp/pulp_npm/actions/runs/662492434

templates/github/.ci/scripts/publish_plugin_pypi.sh.j2
fao89: why draft? It seems ready to me

templates/github/.github/workflows/release.yml.j2
```diff
@@ -127,30 +176,10 @@ jobs:
...
-        run: .ci/scripts/publish_docs.sh tag ${GITHUB_REF##*/}
-      {%- endif %}
-
-      {%- if update_redmine %}
-      - name: Update Redmine
-        run: bash .ci/scripts/update_redmine.sh
-      {%- endif %}
+      - name: {{ "${{ matrix.name }}" }}
+        run: |
+          git checkout "origin/${GITHUB_REF##*/}" -- .ci
```
daviddavis, line 181: It’s probably better to use the branch’s ci scripts rather than the tag’s version of the ci scripts.

templates/github/.ci/scripts/publish_plugin_pypi.sh.j2
```diff
@@ -7,6 +7,14 @@ cd "$(dirname "$(realpath -e "$0")")"/../..
 
 set -euv
 
+export VERSION=$(http pulp/pulp/api/v3/status/ | jq --arg plugin {{ plugin_app_label }} -r '.versions[] | select(.component == $plugin) | .version')
+export response=$(curl --write-out %{http_code} --silent --output /dev/null https://pypi.org/project/pulp-npm-client/$VERSION/)
```
ipanova, line 11: please don't forget to remove hard-coded npm plugin
s/pulp-npm-client/{{ plugin_dash }}-client

daviddavis:
Setting this back to WIP. I have a couple more ideas for improvement.

daviddavis:
Nevermind, the changes are turning out to be quite large. Will follow up with another PR.

CHANGES/8404.feature
```diff
@@ -0,0 +1 @@
+Created a manual trigger for our release process so the process could be rerun in case of failures.
```
mdellweg, line 1: Should this be misc?

daviddavis: I was thinking feature since the audience of the changelog is plugin writers and we want to inform them that the release process is now dispatched manually instead of automatically when they push a tag?

mdellweg: Right this is the plugin_template. Not pulpcore...

templates/github/.github/workflows/release.yml.j2
```diff
@@ -127,30 +176,10 @@ jobs:
...
-
-      {%- if publish_docs_to_pulpprojectdotorg %}
-      - name: Publish docs to pulpproject.org
-        run: .ci/scripts/publish_docs.sh tag ${GITHUB_REF##*/}
-      {%- endif %}
-
-      {%- if update_redmine %}
-      - name: Update Redmine
-        run: bash .ci/scripts/update_redmine.sh
-      {%- endif %}
```
mdellweg, line 153: The "if"s from here should be represented in the matrix above, IMHO.

templates/github/.github/workflows/release.yml.j2
mdellweg: I'm not quite following here. Why?
And which branch do you mean? "2.2" for a "2.2.5" release?

templates/github/.github/workflows/release.yml.j2
daviddavis: Oh yea, I totally forgot. Thanks.

templates/github/.github/workflows/release.yml.j2
daviddavis: This workflow runs against the tag (see [here](https://github.com/pulp/plugin_template/pull/357/files#diff-2804d6731049ff36e9d9c9de50d1541440740aa9fd2479a75714f4e1bfc361ddR132)). It does so in order to build and publish the tag's files to pypi, docs.pulpproject.org, etc.

However, consider the case that you try to release a tag but encounter a problem in the CI scripts. You fix the CI scripts but the workflow will still run against the tag's version of the CI scripts. So the only fix would be to create a new tag with new ci files.

To prevent such a problem, I thought we could check out the .ci scripts from the branch instead.

daviddavis: Maybe I am overthinking things though?

mdellweg: So in case the Release failed, and rerunning does not suffice, but you need to fix CI files, you push them to the release branch and then restart? Maybe this _is_ solving non-obvious corner-case problems. But we should have this troubleshooting method documented somewhere and linked from/to this comment.

daviddavis: I added a comment to the release workflow. I'll also add a section about troubleshooting failed releases to our release guide.

daviddavis:
I decided to revert the publish matrix. I realized that it wasn't necessary to parallelize the publish tasks. It was adding extra complexity. We are checking that the release artifact is published and skipping the step it should run through all publish steps every time it's triggered.

daviddavis:
Ping @ipanova @fao89 @mdellweg any more feedback?

fao89 (approved):
let's try it!

mdellweg (approved):
YOLO

PR #360, opened 2021-03-30
Add docs/_static/ to default .gitignore

quba42:
[noissue]

Since docs/_static/api.json should no longer be commited to source, we
can gitignore the folder.

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -68,6 +68,10 @@ DEPRECATED_FILES = {
         ".travis",
         ".travis.yml",
     ],
+    "docs": [
+        "docs/_static/.this_makes_it_so_git_keeps_the__static_folder",
+        "docs/_static/api.json"
+    ],
 }
 
 
--- a/README.md
+++ b/README.md
@@ -432,8 +432,8 @@ plugin is installed endpoints in the live api docs will be automatically populat
 
 When you run 'make html' command to build the docs, you must have the pulp-api running on
 localhost. The 'make html' command first downloads the OpenAPI schema for the plugin and saves it
-in ``docs/_static/api.json``. You should add this file to git. This file will then provide data
-needed to display the restapi.html page in the root of the built docs.
+in ``docs/_static/api.json``. This file will then provide data needed to display the restapi.html
+page in the root of the built docs.
 
 # Github Actions configuration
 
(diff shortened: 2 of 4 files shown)


mdellweg:
Should we maybe ignore `_static/api.json` and `.git-keep` the folder? I see nothing in the template to create it before calling curl for `api.json`.

quba42:
> Should we maybe ignore `_static/api.json` and `.git-keep` the folder? I see nothing in the template to create it before calling curl for `api.json`.

In the `pulp_deb` repo there is a `docs/_static/.this_makes_it_so_git_keeps_the__static_folder` file that is checked in (not sure how it got there).
Personally I think it would be more elegant to gitignore the entire folder, and then make sure the docs Makefile creates the folder when it needs it. Not sure I am immediately motivated to add that work to this PR though...

quba42:
The latest version will bootstrap a plugin that does not have the `docs/_static` folder.

quba42:
I saw one can also actively "deprecate" the presence of certain files.

Not sure if it is desirable in this case, since the files should only be dropped if one also grabs the makefile changes from `./plugin-template docs PLUGIN_NAME` and I am not sure that that is obvious for template users?

mdellweg:
So you are saying, we should deprecate files selectively by template section.

quba42:
> So you are saying, we should deprecate files selectively by template section.

I guess I am saying: Right now there is a defacto distinction between the `--github` option (which everyone is reapplying all the time), and the other template commands that are not necessarily designed for frequent reapplication.

If I now deprecate files that should be dropped in combination with reapplying something other than `--github` that _might_ trip people up. :wink:

This would be less of an issue if more parts of the template were built for reapplication to existing plugins...

plugin-template
```diff
@@ -65,6 +65,8 @@ DEPRECATED_FILES = [
     ".github/workflows/fips.yml",
     ".travis",
     ".travis.yml",
+    "docs/_static/.this_makes_it_so_git_keeps_the__static_folder",
+    "docs/_static/api.json"
```
mdellweg, line 69:
```suggestion
    # Those files are deprecated from the docs template section. It would be unsafe to delete them without reapplying the latter.
    # "docs/_static/.this_makes_it_so_git_keeps_the__static_folder",
    # "docs/_static/api.json"
```

I agree that it's unsafe today. But it would be a shame to forget about them.

quba42: Now seperated by section.

mdellweg (approved):

bmbouter (approved):
This makes sense to me, thanks!

PR #361, opened 2021-03-30
Properly handle duplicate issues in release script

daviddavis:
It's possible that an issue has two changelog entries which leads to
dupes in the issues_to_close list.

[noissue]

### DIFF
--- a/templates/github/.ci/scripts/release.py.j2
+++ b/templates/github/.ci/scripts/release.py.j2
@@ -76,12 +76,11 @@ with open(f"{plugin_path}/setup.py") as fp:
 release_version = version.replace(".dev", "")
 
 
-issues_to_close = []
+issues_to_close = set()
 for filename in Path(f"{plugin_path}/CHANGES").rglob("*"):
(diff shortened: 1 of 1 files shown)


templates/github/.ci/scripts/release.py.j2
```diff
@@ -76,12 +76,12 @@ with open(f"{plugin_path}/setup.py") as fp:
 release_version = version.replace(".dev", "")
 
 
-issues_to_close = []
+issues_to_close = set()
 for filename in Path(f"{plugin_path}/CHANGES").rglob("*"):
     if filename.stem.isdigit():
         issue = filename.stem
         issue_url = f"{REDMINE_URL}/issues/{issue}.json"
```
daviddavis, line 83: Also, I can't tell what this line is doing. I think it can be removed?

fao89: I think so

fao89 (approved):

mdellweg (approved):

PR #362, opened 2021-03-30
Fix the tag for the publish docs job

daviddavis:
[noissue]


daviddavis:
@quba42 I manually fixed the pulp_deb docs but this fix should fix the problem going forward.

fao89 (approved):

PR #364, opened 2021-04-05
Enable PR checkout from repos outside pulp org

fao89:
https://pulp.plan.io/issues/8510
closes #8510

Ref: https://github.com/pulp/pulp_ansible/pull/551

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,11 @@ You can edit them according to your needs to control subsequent calls to `plugin
 The following settings are stored in `template_config.yml`.
 
 ```bash
-  additional_plugins    A list with additional plugins to be installed on Github Actions.
+  additional_repos    A list with additional repos to be installed on Github Actions.
                         Each item in the list is a dict with the following fields:
                         name: the name of the plugin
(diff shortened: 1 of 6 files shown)


daviddavis (approved):

daviddavis:
We should probably call it `additional_repos` instead of `additional_plugins` but that seems like a job for another day.

daviddavis:
One thought: we should probably update the README to call out these fields for `additional_plugins`.

README.md
```diff
@@ -47,9 +47,11 @@ You can edit them according to your needs to control subsequent calls to `plugin
 The following settings are stored in `template_config.yml`.
 
 ```bash
-  additional_plugins    A list with additional plugins to be installed on Github Actions.
+  additional_repos    A list with additional repos to be installed on Github Actions.
                         Each item in the list is a dict with the following fields:
                         name: the name of the plugin
+                        bindings: boolean whether or not generate bindings
+                        org: github organization
```
daviddavis, line 54: Are these optional? If so, can you add `(optional)` to these lines.

daviddavis (approved):

PR #371, opened 2021-04-14
Add status api fallbacks to the version checks

mdellweg:
On old release branches we still need to perform releases with
pulpcore<3.11 which provides the version info by plugin_snake instead of
app_label.

[noissue]


fao89 (approved):
LGTM

daviddavis (approved):

PR #372, opened 2021-04-14
Remove superfluous line

mdellweg:
[noissue]

Forgot to remove that after inserting the better version.


quba42 (approved):
Does not seem controversial. :stuck_out_tongue_winking_eye:

daviddavis (approved):

PR #374, opened 2021-04-15
Update bootstrap code to use Distribution model

daviddavis:
fixes #8569


templates/bootstrap/plugin_name/app/serializers.py.j2
```diff
@@ -121,10 +123,18 @@ class {{ plugin_camel_short }}DistributionSerializer(platform.PublicationDistrib
...
     class Meta:
-        validators = platform.PublicationDistributionSerializer.Meta.validators + [
+        validators = platform.DistributionSerializer.Meta.validators + [
             myValidator1, myValidator2]
     """
 
+    publication = platform.DetailRelatedField(
+        required=False,
+        help_text=_("Publication to be served"),
+        view_name_pattern=r"publications(-.*/.*)?-detail",
```
mdellweg, line 133:
```suggestion
        view_name="publications-{{ plugin_snake_short }}/{{ plugin_snake_short }}-detail",
```
... or similar

fao89 (approved):
LGTM

templates/bootstrap/plugin_name/app/serializers.py.j2
```diff
@@ -121,10 +123,18 @@ class {{ plugin_camel_short }}DistributionSerializer(platform.PublicationDistrib
...
     """
 
+    publication = platform.DetailRelatedField(
+        required=False,
+        help_text=_("Publication to be served"),
+        view_name="publications-{{ plugin_snake_short }}/{{ plugin_snake_short }}-detail",
+        queryset=models.Publication.objects.exclude(complete=False),
+        allow_null=True,
+    )
+
```
dralley, line 137: Now that the two are combined, we could put both the `publication` and `repository_version` options with some comments about disabling the one you don't need.

daviddavis: The plugin_template as a whole assumes that you're generating a plugin with publications. What we really need is a switch to allow plugin writers to disable publications if they're not using them. But that's a task for another day.

What I can do is add the field commented out with a note to uncomment it and remove `publication` field if the plugin is not using publications. Does that make sense? Also, I think I need to add a repository_version field and a repository field since I believe you can just set repository and have it auto-distribute the latest repo version.

dralley: >What if I add the repository_version field commented out with a note to uncomment it and remove publication field if the plugin is not using publications?

Yup that's exactly what I'm suggesting

>and a repository field since I believe you can just set repository which attaches the latest repo version.

I believe the new core serializer assumes the existence of `repository` so you don't need to deal with it in the plugin.

mdellweg (approved):

PR #377, opened 2021-04-27
Set PULP_URL instead of modifying generate.sh

daviddavis:
Requires https://github.com/pulp/pulp-openapi-generator/pull/44


mdellweg (approved):
LGTM
CI should be green, once the generator pr is merged.

PR #378, opened 2021-04-28
Fix cherry pick of noissue commits

daviddavis:
[noissue]

### DIFF
--- a/templates/github/.ci/scripts/cherrypick.sh.j2
+++ b/templates/github/.ci/scripts/cherrypick.sh.j2
@@ -18,8 +18,13 @@ issue="$2"
 backport="$3"
 commit_message=$(git log --format=%B -n 1 $commit)
 
-echo $commit_message | grep -q -E "(fixes|closes).*#$issue" || \
-  (echo "Error: issue $issue not detected in commit message." && exit 1)
+if ! echo $commit_message | grep -q "\[noissue\]"
+then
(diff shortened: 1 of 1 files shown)


fao89 (approved):
LGTM

mdellweg (approved):
If i read that correctly, when you pick a `[noissue]` commit, `$2` and `$3` are simply ignored in the check, but still used later?
What if you have no issue, but a backport request?
I guess, in the current state, you are advised to amend the commit message?

daviddavis:
> If i read that correctly, when you pick a [noissue] commit, $2 and $3 are simply ignored in the check, but still used later?

That's correct.

> What if you have no issue, but a backport request?

Can you think of such a case?

The issue I dealt with was that the issue essentially had two commits: one noissue and one attached to the issue. So I backported both using the original issue's #.

goosemania:
> > What if you have no issue, but a backport request?
> 
> Can you think of such a case?
> 
> The issue I dealt with was that the issue essentially had two commits: one noissue and one attached to the issue. So I backported both using the original issue's #.

There were backport cases in the past with no issue, however I wouldn't cover that case, and I would break loudly if there is such. If a fix requires a cherry-pick, it definitely deserves an issue in redmine, both the author and the reviewer should make sure that there is one.

daviddavis:
If you don't supply an original issue #, then you can't run this script/you get the help screen. So it does break loudly.

If we do need to backport a [noissue] commit in the future, then we can do so without this script.

So I think we're covered.

PR #379, opened 2021-04-30
Enable https tests

fao89:
closes #8677

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -33,6 +33,7 @@ DEFAULT_SETTINGS = {
     'docker_fixtures': False,
     'docs_test': True,
     'issue_tracker': 'redmine',
+    'pulp_scheme': 'http',
     'plugin_app_label': None,
     'plugin_camel': None,
     'plugin_camel_short': None,
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -113,6 +113,8 @@ VARSYAML
 
 cat >> vars/main.yaml << VARSYAML
 pulp_settings: {{ pulp_settings | tojson }}
+pulp_scheme: {{ pulp_scheme }}
+pulp_container_tag: {{ 'https' if pulp_scheme == 'https' else 'latest' }}
 VARSYAML
 
 {%- if docker_fixtures %}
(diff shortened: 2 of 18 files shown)


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -146,6 +146,13 @@ fi
 ansible-playbook build_container.yaml
 ansible-playbook start_container.yaml
 
+# Hack: adding pulp CA to certifi.where()
+sudo docker cp pulp:/etc/pulp/certs/ca.crt /usr/local/share/ca-certificates/pulp_ca.crt
+CERT=$(python -c 'import certifi; print(certifi.where())')
```
fao89, line 151: This is Mozilla CA bundle, and I'm appending pulp CA to it

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -94,7 +94,7 @@ if [[ "$TEST" = 'bindings' || "$TEST" = 'publish' ]]; then
   gem install --both ./{{ plugin_snake }}_client-0.gem
   cd ..
 {%- endif %}
-  ruby $REPO_ROOT/.ci/assets/bindings/test_bindings.rb
+  SSL_CERT_FILE=/usr/local/share/ca-certificates/pulp_ca.crt ruby $REPO_ROOT/.ci/assets/bindings/test_bindings.rb
```
fao89, line 97: Ruby doesn't use certifi lib, so I had to use an OpenSSL env var to point to the CA

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -146,6 +146,24 @@ fi
 ansible-playbook build_container.yaml
 ansible-playbook start_container.yaml
 
+echo ::group::SSL
+# Copy pulp CA
+sudo docker cp pulp:/etc/pulp/certs/pulp_webserver.crt /usr/local/share/ca-certificates/pulp_webserver.crt
+sudo docker cp pulp:/etc/pulp/certs/pulp_webserver.crt /etc/ssl/certs/pulp_webserver.crt
+sudo docker cp pulp:/etc/pulp/certs/pulp_webserver.key /etc/ssl/private/pulp_webserver.key
```
mdellweg, line 153: I think, we don't need the key here. We should not expose it more than needed.

mdellweg (approved):
Quite a change... But it looks good to me!

fao89:
Since this is approved should we merge it?
About the single container part https://github.com/pulp/pulp-oci-images/pull/73 I really don't know what to do, maybe we should decide on the next CI/CD meeting

mdellweg:
One last thing to consider: If we apply this template to old release branches, will they have a problem?
There is nothing that sticks out to me, however.

daviddavis:
We could open a pr and find out?

fao89:
> One last thing to consider: If we apply this template to old release branches, will they have a problem?
> There is nothing that sticks out to me, however.

since we are deciding to have HTTP and HTTPS tags, I can change this PR for https be optional

daviddavis:
I think that makes sense. Some plugins (like pulp_ansible) required updates to their tests since they do pulp-to-pulp syncs, etc. and these test changes might be hard to backport to older release branches.

I would recommend we default to https being true.

fao89:
Blocked by https://github.com/pulp/pulp-oci-images/pull/73 or https://github.com/pulp/pulp-oci-images/pull/84

daviddavis:
Hey, can you attach this to https://github.com/pulp/plugin_template/issues/403?

plugin-template
```diff
@@ -33,6 +33,7 @@ DEFAULT_SETTINGS = {
     'docker_fixtures': False,
     'docs_test': True,
     'issue_tracker': 'redmine',
+    'pulp_scheme': 'http',
```
fao89, line 36: made `http` default, for now, so I can merge it.
After merging https://github.com/pulp/pulp-oci-images/pull/73 or https://github.com/pulp/pulp-oci-images/pull/84
we can change it to `https`

PR #383, opened 2021-05-10
Update manifest

quba42:
Two small commits: One makes the entries in `MANIFEST.in` use alphabetic order, the second adds the `COPYRIGHT` file to the list.

If `COPYRIGHT` should not be in the `MANIFEST.in` file, please explain why. :wink:

### DIFF
--- a/templates/bootstrap/COPYRIGHT
+++ b/templates/bootstrap/COPYRIGHT
@@ -1,0 +1,10 @@
+Copyright © 2017 Pulp Project developers.
+
+This software is licensed to you under the GNU General Public
+License as published by the Free Software Foundation; either version
+2 of the License (GPLv2) or (at your option) any later version.
+There is NO WARRANTY for this software, express or implied,
+including the implied warranties of MERCHANTABILITY,
+NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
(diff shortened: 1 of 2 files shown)


mdellweg:
The only thing I see, is that there is no `COPYRIGHT` in the bootstrap template.
Funny enough, this repository has one. What _is_ the right policy about having that file here?

I would expect, that the build/release fails, if a file from the Manifest is missing.

quba42:
> The only thing I see, is that there is no `COPYRIGHT` in the bootstrap template.
> Funny enough, this repository has one. What _is_ the right policy about having that file here?

You are right bootstrapping a new plugin from the template does not create a `COPYRIGHT` file, but every pulp repo I checked appears to have one, and at least the ones I have checked have added it to their `MANIFEST.in` file. What is the purpose of having a `COPYRIGHT` file in addition to the `LICENSE` file? Is this even needed?

> I would expect, that the build/release fails, if a file from the Manifest is missing.

Probably. So if every existing plugin has one, maybe it should be added to the plugin template as well?

mdellweg:
> Probably. So if every existing plugin has one, maybe it should be added to the plugin template as well?

I'm about to jump to the same conclusion.

dkliban (approved):

PR #384, opened 2021-05-14
Upgrade tests

fao89:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -17,6 +17,7 @@ from jinja2 import Environment, FileSystemLoader
 
 DEFAULT_SETTINGS = {
     'additional_repos': [],
+    'upgrade_range': [],
     'black': True,
     'check_commit_message': True,
     'check_gettext': True,
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -146,3 +146,89 @@ jobs:
           docker exec pulp ls -latr /etc/yum.repos.d/
           docker exec pulp cat /etc/yum.repos.d/*
           {%- endif %}
+
+{%- if upgrade_range %}
+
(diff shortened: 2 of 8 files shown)


plugin-template
```diff
@@ -17,6 +17,7 @@ from jinja2 import Environment, FileSystemLoader
 
 DEFAULT_SETTINGS = {
     'additional_repos': [],
+    'upgrade_range': [],
```
dkliban, line 20: Please add this variable to the README

fao89: done!

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -146,3 +146,89 @@ jobs:
...
+
+      - name: Install httpie
+        run: |
+          echo ::group::HTTPIE
+          sudo apt-get update -yq
+          sudo -E apt-get -yq --no-install-suggests --no-install-recommends install httpie
+          echo ::endgroup::
+          echo "TEST=upgrade" >> $GITHUB_ENV
+          echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.ci/assets/httpie/" >> $GITHUB_ENV
+          {%- for plugin_branch, version in upgrade_range[0].items() %}
```
dkliban, line 188: Why is the index 0 needed here?

fao89: `upgrade_range` is a list of dicts:
```python
upgrade_range = [
   { "pulpcore_branch": "X.Y", "pulp_file_branch": "X.Y"},
   { "pulpcore_branch": "X.Y", "pulp_file_branch": "X.Y"}
]
```
the keys would repeat, so I took only the first dict keys

dkliban (approved):

PR #389, opened 2021-05-18
Install pulp-smash before plugins

fao89:
`--upgrade --force-reinstall` from pulp-smash may lead to errors

[noissue]


mdellweg:
I really hate pip's approach to dependency resolution. The only clean way i ever found was clumping all package installations into a single pip call.

mdellweg (approved):

PR #390, opened 2021-05-20
Use generic CLI commands on upgrade tests

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -111,8 +111,7 @@ fi
 
 cd pulp-cli
 pip install -e .
-pulp config create --base-url http://pulp --location tests/settings.toml
-sed -i "s/true/false/g" tests/settings.toml
+pulp config create --base-url http://pulp --location tests/settings.toml --no-verify-ssl
 mkdir ~/.config/pulp
 cp tests/settings.toml ~/.config/pulp/settings.toml
 cd ..
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -146,11 +146,12 @@ if [[ "$TEST" == "upgrade" ]]; then
 
   # CLI commands to display plugin versions and content data
   pulp status
-  {%- if plugin_name != 'pulpcore' %}
-  pulp {{ plugin_app_label }} content list
-  {%- else %}
-  pulp file content list
-  {%- endif %}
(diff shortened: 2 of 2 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -111,8 +111,7 @@ fi
 
 cd pulp-cli
 pip install -e .
-pulp config create --base-url http://pulp --location tests/settings.toml
-sed -i "s/true/false/g" tests/settings.toml
+pulp config create --base-url http://pulp --location tests/settings.toml --no-verify-ssl
```
mdellweg, line 114: :100:

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -146,11 +146,8 @@ if [[ "$TEST" == "upgrade" ]]; then
 
   # CLI commands to display plugin versions and content data
   pulp status
-  {%- if plugin_name != 'pulpcore' %}
-  pulp {{ plugin_app_label }} content list
-  {%- else %}
-  pulp file content list
-  {%- endif %}
+  pulp show --href /pulp/api/v3/content/
```
mdellweg, line 149: This is not a supported command.
show is meant to be used with detail views. And we will probably add some validation here that will break this usage.

fao89: it worked with pulp_ansible, what is the best generic command that I can put here?
See previous discussion: https://github.com/pulp/pulp_ansible/pull/576#discussion_r636197472

mdellweg: Well this is nothing more than the plain `http $HOST/pulp/api/v3/content/`. And also it's not supported and may break in the future. To list content, you'd need to specify the type like `pulp file content list` (where "content" is the content type unit).
Also a generic content list command is just a few lines of CLI code away.

Quite honestly i'm mostly wondering what this command is supposed to test. Calling out to the generic content endpoint and not even looking at the result seems rather pointless to me.

fao89: it is for debugging purposes, I wanted to answer the following questions:
`pulp status` - are pulpcore and plugins really upgraded?
`pulp content list` - did we keep the data after the migration?

mdellweg: https://github.com/pulp/pulp-cli/pull/254

Here you go.

mdellweg (approved):

PR #393, opened 2021-05-27
Don't run CI on branch push

dralley:
The marginal utility of running tests on branches immediately after
merging a PR where the tests were already run is not zero, but it is
very low. Let's reduce our CI consumption by getting rid of the
duplication. Nightly builds are probably sufficient for the small
likelihood of post-merge errors.

[noissue]

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -5,9 +5,6 @@ on:
   pull_request:
     branches:
       - '*'
-  push:
-    branches:
-      - '*'
 
 jobs:
 


daviddavis:
Thanks for this PR. I don't have a strong opinion about keeping or disabling the CI for pushes but I agree that it's of marginal benefit to us. 

@pulp/continuous-integration any thoughts?

bmbouter:
:+1:

fao89 (commented):
we can always give it a try and revert if needed, so I'm +1 to merge it

mdellweg (approved):
I'm approving this, also because a failed CI on a push to the main branch does not even have an immediate consequence.
One last thought: We could turn it into a manual trigger. So we could identify whether a problem on a PR is really the PR, or something on the main branch went wrong without waiting for the nightly.

PR #395, opened 2021-06-15
Handles plugins with dashes in names instead of underscores

dkliban:
This change also properly handles plugins that need to be tested with more than
one other plugin.

[noissue]


mdellweg (approved):
LGTM

mdellweg:
Maybe you want to change the name of the test template plugin to "pulp_cat-dog" ?

PR #397, opened 2021-06-17
Include git reference in the template

mdellweg:
This may help to regenerate the CI on old branches if applying the latest is not an option.


daviddavis:
This is a great idea 👍

quba42:
I have been manually writing these into commit messages whenever I remembered to do so. This is better! :+1:

fao89 (approved):
🎉

mdellweg:
This is covering the template_config, github and docs section. I no one asks for more places, i will merge on monday.

quba42 (approved):

PR #399, opened 2021-06-17
Fix pypi publishing for clients whose name != snake.

ggainey:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/publish_client_pypi.sh.j2
+++ b/templates/github/.github/workflows/scripts/publish_client_pypi.sh.j2
@@ -9,7 +9,7 @@ cd "$(dirname "$(realpath -e "$0")")"/../../..
 
 pip install twine
 
-export VERSION=$(ls dist | sed -rn 's/{{ plugin_name }}-client-(.*)\.tar.gz/\1/p')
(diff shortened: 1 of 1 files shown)


dralley (approved):

templates/github/.github/workflows/scripts/publish_client_pypi.sh.j2
```diff
@@ -9,11 +9,12 @@ cd "$(dirname "$(realpath -e "$0")")"/../../..
 
 pip install twine
 
-export VERSION=$(ls dist | sed -rn 's/{{ plugin_name }}-client-(.*)\.tar.gz/\1/p')
+export VERSION=$(ls dist | sed -rn 's/{{ plugin_snake }}-client-(.*)\.tar.gz/\1/p')
 
 if [[ -z "$VERSION" ]]; then
   echo "No client package found."
-  exit
+  # Not-publishing should be fatal
```
dkliban, line 16: This should not be fatal. The package is not found if it had been published before. However, we should probably be checking with PyPI to ocnfirm that a package does exist. Such a check happens earlier in the workflow.

ggainey: Fair enough - removed

dkliban (approved):

PR #414, opened 2021-06-21
Remove redmine step from release guide

daviddavis:
[noissue]


daviddavis:
This makes me think we should maybe think about creating milestones in plugin_template and track PRs/issues. I leave this up to some future person.

fao89 (approved):

PR #421, opened 2021-06-23
Also show flake8 version in CI pipeline

quba42:
[noissue]

We already have this for black. It can help understand why a locally installed flake8 did not alert to the same things as the one in the CI.


mdellweg (approved):
Nice!
Seems like this also shows the flake plugins installed.

Issue #433, opened 2021-07-02
release workflow should allow user to specify custom post_before_script.sh


dkliban:
In cases where docs need to be fixed after a release, a post_before_script.sh can be used to checkout the correct docs. The release workflow needs to accept an optional parameter that can be used to populate the content of the post_before_script.sh.

PR #434, opened 2021-07-02
Adds an optional 'before_script' parameter to the release worklow.

dkliban:
This parameter should be used when re-running a release workflow in
order to publish updated docs.

fixes: #433


bmbouter (approved):
Thanks!

PR #436, opened 2021-07-06
Update paths to python 3.8

daviddavis:
[noissue]


fao89 (approved):
LGTM, but I think we need a changelog entry

templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -10,7 +10,7 @@ ADD {{ item.source }} {{ item.source }}
 # Install python packages
 # Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
-RUN pip3 install \
+RUN pip3 install packaging \
```
daviddavis, line 13: Real fix is here: https://github.com/pulp/pulp-oci-images/pull/104

PR #437, opened 2021-07-07
Adding "after failure" to nightly

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -136,7 +136,7 @@ jobs:
         shell: bash
 
       - name: After failure
-        if: {{ "${{ failure() }}" }} 
+        if: failure()
         run: |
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -136,7 +136,7 @@ jobs:
         shell: bash
 
       - name: After failure
-        if: {{ "${{ failure() }}" }} 
+        if: failure()
```
daviddavis, line 139: This seems right to me but it also seems that this was recently changed?

https://github.com/pulp/plugin_template/pull/435

I guess ttereshc mistakenly thought the after failure step wasn't being run for nightly because of the formatting?

fao89: That was my guess also

daviddavis (approved):

PR #442, opened 2021-07-12
Adjustments for galaxy

fao89:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -19,6 +19,7 @@ DEFAULT_SETTINGS = {
     'additional_repos': [],
     'upgrade_range': [],
     'black': True,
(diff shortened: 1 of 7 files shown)


plugin-template
```diff
@@ -19,6 +19,7 @@ DEFAULT_SETTINGS = {
     'additional_repos': [],
     'upgrade_range': [],
     'black': True,
+    'flake8': True,
```
daviddavis, line 22: Any reason why not to expose this in the README and bootstrap code?

fao89: I just forgot

daviddavis (approved):

PR #445, opened 2021-07-13
Commit message

pavelpicka:
Commit message must be extracted before any git checkout.
If not, before_install script doesn't recognize `Required PR:` correctly for upgrade test, as it is working with wrong commit message.


fao89 (approved):
Nice catch!

dkliban (approved):

PR #447, opened 2021-07-15
Make https default

fao89:
Depends on https://github.com/pulp/pulp-oci-images/pull/105


daviddavis:
Can you add a changelog entry? I'm guessing most plugins are using http right now and this change will cause them to switch to https and I think greater visibility for plugin writers would help.

And optionally, I'd probably break the non-https changes to requirements.txt and setup.py into a separate PR (or commit).

PR #455, opened 2021-07-20
Fix branch refs for update ci job

daviddavis:
[noissue]


daviddavis:
@fao89 is this correct?

fao89 (commented):
I believe so

fao89 (approved):

PR #456, opened 2021-07-20
Update the template to match the latest version of pulpcore

lubosmj:


CHANGES.rst
```diff
@@ -7,7 +7,7 @@ Changelog
     file is managed by towncrier. You *may* edit previous change logs to
     fix problems like typo corrections or such.
     To add a new change log entry, please see
-    https://docs.pulpproject.org/en/3.0/nightly/contributing/git.html#changelog-update
+    https://docs.pulpproject.org/pulpcore/contributing/git.html#changelog-update
```
fao89, line 10: the link you provided is the latest released,
this is the one that is built nightly:
https://docs.pulpproject.org/pulpcore/en/master/nightly/contributing/git.html#changelog-update

templates/bootstrap/plugin_name/app/viewsets.py.j2
```diff
@@ -60,7 +60,7 @@ class {{ plugin_camel_short }}ContentViewSet(core.ContentViewSet):
         "Artifacts" need to be popped off and saved indpendently, as they are not actually part
         of the Content model.
         """
-        raise NotImplementedError("FIXME")
```
fao89, line 63: I think this is more explicit

templates/bootstrap/plugin_name/app/tasks/synchronizing.py.j2
```diff
@@ -93,4 +93,4 @@ class {{ plugin_camel_short }}FirstStage(Stage):
         Args:
             path: Path to the metadata file
         """
-        pass
+        return []
```
fao89, line 96: not sure if we should return it, what was the error you saw here? I'm afraid returning an empty list will make harder for people to find where they need to implement

templates/bootstrap/plugin_name/app/viewsets.py.j2
```diff
@@ -160,11 +160,16 @@ class {{ plugin_camel_short }}RepositoryViewSet(core.RepositoryViewSet, ModifyRe
...
 
         result = dispatch(
             tasks.synchronize,
             [repository, remote],
-            kwargs={"remote_pk": str(remote.pk), "repository_pk": str(repository.pk)},
+            kwargs={
+                "remote_pk": str(remote.pk),
+                "repository_pk": str(repository.pk),
+                "mirror": mirror,
+            },
```
fao89, line 172: Nice!

templates/test/dev_requirements.txt.j2
```diff
@@ -1,3 +1,4 @@
+check-manifest
```
fao89, line 1: sweet!

CHANGES.rst
lubosmj: Yes, I know that this is for the latest version released. I thought that `/en/3.0/nightly/` should be removed as a whole since it was placed everywhere where we talk about `pulpcore`. Was it on purpose?

templates/bootstrap/plugin_name/app/viewsets.py.j2
lubosmj: After this change, users who initialized a new Pulp plugin will not get an HTTP 500 `Internal Server Error` response. Instead, they will receive an HTTP 501 `Not Implemented` response.

templates/bootstrap/plugin_name/app/tasks/synchronizing.py.j2
lubosmj: A similar case to the previous comment. Users will not need to dig in the journals to see the error message and why Pulp even raised an error. These two changes make a plugin consistent with the documentation, meaning that the plugin will somehow work (as specified in the docs), but still provide reasonable initial behaviour where not all of the features are implemented.

lubosmj: The error message was basically - `None object is not iterable` (or something like that), since it returned `None` (https://github.com/pulp/plugin_template/blob/74c9bb33c41051fe78321ff255413c1d7a31af9f/templates/bootstrap/plugin_name/app/tasks/synchronizing.py.j2#L76).

CHANGES.rst
fao89: you are right about removing it since the link is broken, 
I believe the change would be
```
/3.0/master
```
nightly build is better because the documentation could be on master and not on the latest release

templates/bootstrap/plugin_name/app/tasks/synchronizing.py.j2
fao89: got it! I was worried it would make it harder to find why it errored, but you are right!

templates/bootstrap/plugin_name/app/viewsets.py.j2
fao89: you are right, it seems better

fao89 (approved):
overall I agree with your points, I just think the documentation needs to point to /en/master/nightly (3.0/nightly was broken indeed). But I checked the links here and they usually don't change a lot, so I'm approving

PR #458, opened 2021-07-22
Lint plugin-template

daviddavis:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -17,52 +17,52 @@ from jinja2 import Environment, FileSystemLoader
 
 
 DEFAULT_SETTINGS = {
-    'additional_repos': [],
-    'upgrade_range': [],
-    'black': True,
-    'flake8': True,
-    'check_commit_message': True,
-    'check_gettext': True,
-    'check_manifest': True,
-    'check_openapi_schema': True,
-    'check_stray_pulpcore_imports': True,
-    'coverage': False,
-    'deploy_client_to_pypi': True,
-    'deploy_client_to_rubygems': True,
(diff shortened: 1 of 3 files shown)


plugin-template
```diff
@@ -345,27 +415,34 @@ def check_for_deprecated_files(plugin_root_dir, sections):
...
+    config["redmine_project"] = None
+    config["update_redmine"] = False
+    config.pop("config")
+    config["config"] = config.copy()
     write_template_section(config, "generate_config", plugin_root_dir, verbose=verbose)
     write_template_section(config, "github", plugin_root_dir, verbose=verbose)
     print("Updating pyproject.toml to point to GitHub issues\n")
     search_string = "pulp.plan.io"
-    replace_string = f"github.com\/pulp\/{name}"
+    replace_string = f"github.com/pulp/{name}"
```
daviddavis, line 435: flake8 alerted on this change and I'm assuming the '\' aren't supposed to be there?

```
>>> name = "catdog"
>>> f"github.com\/pulp\/{name}"
'github.com\\/pulp\\/catdog'
```

daviddavis: @gerrod3 is the above what you expect? It looks like python interprets \/ as \\\/ in f-strings.

Maybe I should just convert it to a normal string and use `format()`?

gerrod3: You can try. I have no idea if it'll work. I think the problem is that the string you pass in has to be a sed format string, but python does weird things with the slashes and that's why I had an escape slash there.

mdellweg (approved):

PR #459, opened 2021-07-23
Allow plugins to configure python_version

### LINKED ISSUE #457: Allow core and plugins to specify which Python plugin to test against
The CI uses the pulp_ci_centos image which already has a version of Python installed. Currently, that's Python 3.8. However, we're still releasing older versions of pulpcore and plugins that support Python 3.6 and we need a way to have our CI test against Python 3.6.

daviddavis:
fixes #457

### DIFF
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -90,7 +90,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.8"
+          python-version: "{{ python_version }}"
 
       {%- if test_bindings or deploy_client_to_rubygems %}
       - uses: actions/setup-ruby@v1
--- a/CHANGES/457.feature
+++ b/CHANGES/457.feature
@@ -1,0 +1,1 @@
+Added python_version option.
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ The following settings are stored in `template_config.yml`.
                         is required unless deploy_client_to_pypi and deploy_daily_client_to_pypi
                         and deploy_to_pypi are specified.
 
+  python_version        Python version to use in the CI. Currently only 3.6 and 3.8 are supported.
+
   redmine_project       A string that corresponds to the redmine identifier for the repo's project.
                         This is used during commit validation to make sure the commit is attached to
                         an issue in the correct project.
--- a/plugin-template
+++ b/plugin-template
@@ -53,6 +53,7 @@ DEFAULT_SETTINGS = {
     'pulpprojectdotorg_key_id': None,
     'pydocstyle': True,
     'pypi_username': None,
+    'python_version': "3.8",
     'redmine_project': None,
     'release_user': 'pulpbot',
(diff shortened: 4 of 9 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -68,12 +72,15 @@ then
...
+  export CI_BASE_IMAGE=$(echo $COMMIT_MSG | sed -n -e 's/.*CI Base Image:\s*\([-_/[:alnum:]]*:[-_[:alnum:]]*\).*/ci_base: "\1"/p')
+  if [[ ! -z "$CI_BASE_IMAGE" ]]; then
+    sed -i '/ci_base:/d' .ci/ansible/vars/main.yml
+    echo $CI_BASE_IMAGE >> .ci/ansible/vars/main.yaml
+  fi
 else
   {%- for repository in ["pulpcore", "pulp-smash", "pulp-openapi-generator", "pulp-cli"] + additional_repos | map(attribute="name") | list %}
   export {{ repository | upper | replace("-", "_") }}_PR_NUMBER=
 {%- endfor %}
-  export CI_BASE_IMAGE=
```
daviddavis, line 76: Not sure what this was for so I repurposed it.

dralley (approved):

daviddavis:
I could use a review from @pulp/continuous-integration

templates/github/.github/workflows/release.yml.j2
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.8"
+          python-version: "{{ python_version }}"
```
dralley, line 35: We only need to be concerned with having the CI run on Python 3.6 - the release scripts should run on either 3.8 or "latest" because they break on 3.6

https://github.com/pulp/pulp_rpm/runs/3147430536?check_suite_focus=true#step:7:82

dralley (changes requested):
See comment

templates/github/.github/workflows/release.yml.j2
dralley:
```
diff --git a/templates/github/.github/workflows/release.yml.j2 b/templates/github/.github/workflows/release.yml.j2
index 9e4e16e..4e1616a 100644
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -32,7 +32,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "{{ python_version }}"
+          python-version: 3.8
 
       - name: Install python dependencies
         run: |
@@ -211,7 +211,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "{{ python_version }}"
+          python-version: "3.8"
 
       - uses: actions/setup-ruby@v1
         with:

```

dralley (approved):

mdellweg (approved):
LGTM

PR #460, opened 2021-07-27
Fixing update ci files workflow

daviddavis:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/update_ci.yml.j2
+++ b/templates/github/.github/workflows/update_ci.yml.j2
@@ -32,14 +32,13 @@ jobs:
       - name: Create Pull Request for CI files
         uses: peter-evans/create-pull-request@v3
         with:
+          token: {{ '${{ secrets.RELEASE_TOKEN }}' }}
           committer: {{  'pulpbot <pulp-infra@redhat.com>' }}
           author: {{  'pulpbot <pulp-infra@redhat.com>' }}
-          branch: ${GITHUB_REF#refs/heads/}-update-ci-files
-          base: ${GITHUB_REF#refs/heads/}
-          title: 'Update CI for ${GITHUB_REF#refs/heads/}'
+          title: 'Update CI files'
           body: '[noissue]'
           commit-message: |
-            Update CI files for ${GITHUB_REF#refs/heads/}
+            Update CI files
 
             [noissue]
           delete-branch: true


templates/github/.github/workflows/update_ci.yml.j2
```diff
@@ -34,12 +34,12 @@ jobs:
         with:
           committer: {{  'pulpbot <pulp-infra@redhat.com>' }}
           author: {{  'pulpbot <pulp-infra@redhat.com>' }}
-          branch: ${GITHUB_REF#refs/heads/}-update-ci-files
-          base: ${GITHUB_REF#refs/heads/}
-          title: 'Update CI for ${GITHUB_REF#refs/heads/}'
+          branch: {{ plugin_default_branch }}-update-ci-files
+          base: {{ plugin_default_branch }}
+          title: 'Update CI files'
```
mdellweg, line 39: Was this supposed to update the files on release branches too?

mdellweg (approved):

templates/github/.github/workflows/update_ci.yml.j2
daviddavis: Eventually I am hoping to handle that. But for now, it just handles `plugin_default_branch`.

daviddavis: Ug yes, good catch. I suppose I need to find a way to handle that. The problem is that `${GITHUB_REF#refs/heads/}` doesn't resolve here. I suppose I need to ditch this `peter-evans/create-pull-request@v3` and write a script to create a PR.

fao89: I don't think you need to ditch it, it works here:
https://github.com/pulp/plugin_template/blob/master/templates/github/.github/workflows/create-branch.yml.j2#L52-L74
I believe you need to create the branch first, and 
```
git checkout -b $BASE_BRANCH-update-ci-files
echo "BASE_BRANCH=some logic before" >> $GITHUB_ENV
```

templates/github/.github/workflows/update_ci.yml.j2
```diff
@@ -34,12 +34,10 @@ jobs:
         with:
           committer: {{  'pulpbot <pulp-infra@redhat.com>' }}
           author: {{  'pulpbot <pulp-infra@redhat.com>' }}
-          branch: ${GITHUB_REF#refs/heads/}-update-ci-files
-          base: ${GITHUB_REF#refs/heads/}
```
daviddavis, line 38: I read through [the docs](https://github.com/peter-evans/create-pull-request#action-inputs) and it looks like if I omit the inputs it'll do what I want by automatically creating a new branch and using the workflow's base branch.

daviddavis:
Looks like the main problem now is that the RELEASE_TOKEN secret doesn't have enough permissions to edit the workflow files.

Will follow up with @RCMariko next week to get a new token with the needed permissions.

dralley (approved):

daviddavis:
I'm going to go ahead and merge this so it will propagate to repos and then I will fix the token next week.

PR #461, opened 2021-07-27
Fix nginx snippets

fao89:
It was breaking pulp-container:
https://github.com/pulp/pulp_container/pull/363#issuecomment-887584318
but weirdly it didn't break pulp-ansible


fao89:
Tested here:
https://github.com/fao89/pulp_container/commit/b9dd9bd996ab4bc6b8583da5e052ce90e772007a
https://github.com/fao89/pulp_container/runs/3174368735?check_suite_focus=true

daviddavis (approved):

PR #465, opened 2021-08-03
Allow custom workflow name

rochacbruno:
I am not sure if there is a need for using `Pulp CI` on every plugin but here is a proposal.

Right now the workflow in a plugin is named `Pulp CI` this changes that
label to be `Plugin CI` e.g: `Galaxy CI`

Before

Screenshot_2021-08-03_21-35-55

After
Screenshot_2021-08-03_21-36-22

[noissue]


daviddavis:
Can you update the nightly job name too?

rochacbruno:
@daviddavis done.

fao89 (approved):

PR #466, opened 2021-08-03
Unpin deps on upgrade tests

fao89:
[noissue]


mdellweg:
I am in agreement to remove this, as it will run the tests with whatever packages will actually be installed at a users site.
My only question: Why do we have it? Does it help to test that our minimal requirements are still valid? Does it provide a more stable CI? Or was it an undocumented workaround for some strange situation?

mdellweg (approved):

PR #468, opened 2021-08-04
Display failed tasks on failure

fao89:
[noissue]

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -135,6 +135,7 @@ jobs:
         run: |
           echo "Need to debug? Please check: https://github.com/marketplace/actions/debugging-with-tmate"
           http --timeout 30 --check-status --pretty format --print hb https://pulp/pulp/api/v3/status/ || true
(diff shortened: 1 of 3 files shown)


.github/workflows/bootstrap.yaml
```diff
@@ -135,6 +135,7 @@ jobs:
         run: |
           echo "Need to debug? Please check: https://github.com/marketplace/actions/debugging-with-tmate"
           http --timeout 30 --check-status --pretty format --print hb https://pulp/pulp/api/v3/status/ || true
+          http --timeout 30 --check-status --pretty format --print hb https://pulp/pulp/api/v3/tasks/?state=failed || true
```
mdellweg, line 138: I would rather see individual tests report their unexpectedly failing tasks. Because this will also show the expectedly failing tasks, and i assume, there are quite a few.
WDYT?

fao89: this is what motivated me:
https://github.com/pulp/pulp_rpm/runs/3240964693?check_suite_focus=true
only now I realize I misread the logs and I could see the failed task there

PR #471, opened 2021-08-17
Use GITHUB_TOKEN in validate_commit_messages

mdellweg:
[noissue]


mdellweg:
I introduced this, because i hit a rate limit. XD

daviddavis (approved):

PR #474, opened 2021-08-23
Minor improvements on plugin-template

fao89:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -306,6 +306,8 @@ def write_template_section(config, name, plugin_root_dir, verbose=False):
     files_copied = 0
     gitref = subprocess.check_output(["git", "describe", "--dirty"]).decode().strip()
     for relative_path in generate_relative_path_set(section_template_dir):
+        if config["issue_tracker"] != "redmine" and "redmine" in relative_path:
+            continue
         destination_relative_path = relative_path.replace("plugin_name", config["plugin_snake"])
         necessary_dir_structure = os.path.dirname(
             os.path.join(plugin_root_dir, destination_relative_path)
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/release.yml.j2
```diff
@@ -32,7 +32,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.8"
+          python-version: "{{ python_version }}"
```
mdellweg, line 35: I think i remember, we discussed this. And the release workflow does not run any pulp code so we opted out from making the python version flexible.
Am i right here? @daviddavis

daviddavis: That's basically correct except I believe there was release Python code that didn't wasn't compatible with older versions of Python.

daviddavis: The release code does use `python_version` when running tests:

https://github.com/pulp/plugin_template/pull/474/files#diff-2804d6731049ff36e9d9c9de50d1541440740aa9fd2479a75714f4e1bfc361ddL93

templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.8"
+          python-version: "{{ python_version }}"
```
daviddavis, line 30: This workflow doesn't run Pulp at all so I don't see any benefit to changing this to `python_version` instead of 3.8.

daviddavis (approved):

PR #477, opened 2021-08-25
Test OpenAPI schema view

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -34,6 +34,8 @@ if [[ "$TEST" = "docs" ]]; then
   cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
   cmd_prefix bash -c "python3 /tmp/schema.py"
   cmd_prefix bash -c "pulpcore-manager spectacular --file pulp_schema.yml --validate"
+  echo "Testing OpenAPI schema view..."
+  http $PULP_URL/pulp/api/v3/
 {%- endif %}
 
   if [ -f $POST_DOCS_TEST ]; then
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -34,6 +34,8 @@ if [[ "$TEST" = "docs" ]]; then
   cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
   cmd_prefix bash -c "python3 /tmp/schema.py"
   cmd_prefix bash -c "pulpcore-manager spectacular --file pulp_schema.yml --validate"
+  echo "Testing OpenAPI schema view..."
+  http $PULP_URL/pulp/api/v3/
```
mdellweg, line 38: What's the reason for this? We check it transitively in at least 2 places, i believe:
- when generating the client bindings
- when running any cli command

Can we at least ensure that the answer is valid json? Maybe open a `::group::` and even paste the schema to the gha logs for debugging. Or save it as an artifact for that workflow for post mortem analysis.

fao89: actually, these are different views, the ones you mentioned are from:
`/pulp/api/v3/docs/api.json?bindings&plugin=<<plugin>>`
which corresponds to https://github.com/pulp/pulpcore/blob/master/pulpcore/app/urls.py#L151-L157
`/pulp/api/v3/` corresponds to:
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/urls.py#L179-L181

mdellweg: Oh right. Shouldn't we remove that to begin with? Is it (useful for)/(used by) anything?

fao89: I don't know, let's discuss at the next open floor

fao89:
closing in favor of https://github.com/pulp/pulpcore/pull/1593

PR #478, opened 2021-08-25
Using Macros for repetitive blocks

rochacbruno:
- Using macros to avoid repetition and allow modularization (easy to
  crate new jobs)
- All scripts runs with the same envvars for consistency

[noissue]


daviddavis (approved):
This is great :+1:

PR #483, opened 2021-08-30
Fix nightly ruby client

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -1,14 +1,14 @@
 {% include 'header.j2' %}
-{% from 'macros.j2' import 
-  set_env_vars, 
-  checkout, 
-  setup_python, 
+{% from 'macros.j2' import
+  set_env_vars,
+  checkout,
+  setup_python,
   setup_ruby,
-  install_httpie, 
-  after_failure, 
-  run_script, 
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -112,7 +112,7 @@ jobs:
 
       {{ run_script(name="Install Python client", file="install_python_client.sh", withenv=False) | indent(6) }}
 
-      {{ run_script(name="Install Ruby client", file="install_ruby_client.sh", withenv=False, condition="${{ env.TEST == 'bindings' || env.TEST == 'generate-bindings' }}") | indent(6) }}
```
fao89, line 115: we weren't installing the ruby bindings

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -1,14 +1,14 @@
 {% include 'header.j2' %}
-{% from 'macros.j2' import
```
fao89, line 2: my editor doesn't like trailing spaces

daviddavis (approved):

templates/github/.github/workflows/nightly.yml.j2
mdellweg: I don't either.

PR #484, opened 2021-08-30
Fix ruby client download

fao89:
[noissue]


ggainey (approved):
Much better - lgtm!

daviddavis (approved):

mdellweg (approved):
That looks correct to me.

PR #485, opened 2021-08-30
Make DEBUG=True on CI jobs

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -166,6 +166,8 @@ then
   exit $s
 fi
 
+sed -i -e 's/DEBUG = False/DEBUG = True/' pulpcore/pulpcore/app/settings.py
+
 cd {{ plugin_name }}
 
 if [ -f $POST_BEFORE_INSTALL ]; then


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -166,6 +166,8 @@ then
   exit $s
 fi
 
+sed -i "s/DEBUG = False/DEBUG = True/" pulpcore/pulpcore/app/settings.py
```
mdellweg, line 169:
```suggestion
sed -i -e 's/DEBUG = False/DEBUG = True/' pulpcore/pulpcore/app/settings.py
```
Just a matter of personal style, but i like to explicit about the "expression" with sed. Also using single quotes usually keeps you from close to crazy shell escaping (a good habit even if not applicable here).

mdellweg: Needless to say, this is exactly what it is labelled: a suggestion

daviddavis:
I'm curious about what benefit enabling DEBUG will bring? Extra logging?

fao89:
> I'm curious about what benefit enabling DEBUG will bring? Extra logging?

yes, it helped me debug https://github.com/pulp/pulpcore/pull/1588

mdellweg (approved):

PR #488, opened 2021-08-31
Fix update ci script

rochacbruno:
runs on galaxy_ng was failing because PR was opened using [noissue] instead of No-Issue

example: https://github.com/ansible/galaxy_ng/pull/910/commits


fao89 (approved):
LGTM

PR #490, opened 2021-09-01
Skip Update CI PR if only template_gitref has changed.

rochacbruno:
When template update has no effect on workflows there is a
change on template_gitref and then useless PRs are opened.

[noissue]


fao89 (approved):
LGTM!

PR #492, opened 2021-09-03
Fixed some links to pulp_file

daviddavis:
[noissue]


fao89 (approved):
I think we should have a target date to rename all master branches to main

daviddavis:
> I think we should have a target date to rename all master branches to main

Yea, I've been doing it repo by repo. The big obstacle will be renaming pulpcore's master since lots of template_config.yml reference pulpcore master (e.g. https://github.com/pulp/pulp_file/blob/1.2/template_config.yml#L34).

PR #494, opened 2021-09-09
Add a single commit check workflow

daviddavis:
[noissue]

### DIFF
--- a/.github/workflows/check_commits.yaml
+++ b/.github/workflows/check_commits.yaml
@@ -2,6 +2,20 @@
 name: Check Commit
 on: pull_request
 jobs:
+  single_commit:
+    name: Assert single commit
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 30
+    - name: Checkout master
+      run: git fetch origin master
+    - name: create local master branch
+      run: git branch master origin/master
+    - name: Commit Count Check
+      run: test `git log  --oneline --no-merges HEAD ^master  | wc -l ` = 1
+    runs-on: ubuntu-latest
   test:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


daviddavis:
Stolen from Katello: https://github.com/Katello/katello/blob/master/.github/workflows/pull_request.yml

I'm imagining that this will be a non-blocking test. In other words, it won't prevent merges. It will just help to warn contributors/PR reviewers that there's more than one commit.

bmbouter:
Awesome! Will this be enabled by default or do plugins need to opt-in? I'm +1 to enabling by default and opting out, but I'm ok w/ whatever you decide.

daviddavis:
Thanks I forgot to set a default. I'm imagining it'll be enabled by default too.

bmbouter (approved):
I kind of liked the discourse announcement part. :) Thanks for doing this!

daviddavis:
> I kind of liked the discourse announcement part. :) Thanks for doing this!

Ha yea, that was a WIP I didn't mean to commit yet. Eventually I want to automatically have the release automation automatically post the announcement to discourse.

rochacbruno (approved):
Nice! I will set it to false on galaxy_ng on next ci update.

We allow multiple commits because we enforce Squash and Merge.

quba42:
I think I will also opt out pulp_deb. The PR volume is not so high that I feel the need to enforce single commits via automation.

mdellweg:
Can this check be used to set/delete a "multiple commit" label on the PR, instead of presenting a failed test?

daviddavis:
I think we could if we used a `pull_request_target` workflow. I personally like having a failing job as it's more obvious and shows near the merge button. But if you think we ought to use a label instead, I'd say open an issue, we can discuss more and I can look into it.

PR #497, opened 2021-09-16
Inject environment vars into pulp container

mdellweg:
[noissue]

~This can help to e.g. inject `DJANGO_ALLOW_ASYNC_UNSAFE` into the pulp container.~
Edit: this did not work out for this specific variable. So i am unsure, if this actually worked.


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -64,6 +64,9 @@ services:
     image: "pulp:${TAG}"
     volumes:
       - ./settings:/etc/pulp
+{%- if pulp_env %}
+    env: {{ pulp_env | tojson }}
+{%- endif %}
```
fao89, line 69: sweet!

rochacbruno:
> This can help to e.g. inject DJANGO_ALLOW_ASYNC_UNSAFE into the pulp container.
> Edit: this did not work out for this specific variable. So i am unsure, if this actually worked.

@mdellweg have you prefixed the env var? to add `settings.DJANGO_ALLOW_ASYNC_UNSAFE` you have to `export PULP_DJANGO_ALLOW_ASYNC_UNSAFE`

mdellweg:
@rochacbruno No, this is really about the environment variable and not an injection to settings. And the way it is handled here https://github.com/pulp/pulpcore/pull/1622/files is renders this approach unusable.
I am going to close this one. Feel free to revive, once you want to inject other env vars.

PR #498, opened 2021-09-16
Pin amazon.aws collection in before_install

mdellweg:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -158,7 +158,7 @@ pip install docker netaddr boto3 ansible
 
 for i in {1..3}
 do
-  ansible-galaxy collection install amazon.aws && s=0 && break || s=$? && sleep 3
+  ansible-galaxy collection install "amazon.aws:1.5.0" && s=0 && break || s=$? && sleep 3
 done
 if [[ $s -gt 0 ]]
 then


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -158,7 +158,7 @@ pip install docker netaddr boto3 ansible
 
 for i in {1..3}
 do
-  ansible-galaxy collection install amazon.aws && s=0 && break || s=$? && sleep 3
+  ansible-galaxy collection install "amazon.aws:1.5.0" && s=0 && break || s=$? && sleep 3
```
fao89, line 161: it seems they reverted your fix!

fao89 (approved):
thank you!

fao89 (approved):

templates/github/.github/workflows/scripts/before_install.sh.j2
mdellweg: Damn it we are back where we started?

fao89: not sure, but I believe so

fao89:
FYI @rochacbruno

templates/github/.github/workflows/scripts/before_install.sh.j2
mdellweg: I don't see it reverted. This may actually be a new bug.

PR #499, opened 2021-09-16
Add tasking_allow_async_unsafe flag

mdellweg:
Plugins that reviewed their tasks to be async safe should disable this.

[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -168,6 +168,11 @@ fi
 
 sed -i -e 's/DEBUG = False/DEBUG = True/' pulpcore/pulpcore/app/settings.py
 
+{% if not tasking_allow_async_unsafe -%}
+# Patch DJANGO_ALLOW_ASYNC_UNSAFE out of the pulpcore tasking_system
+# Don't let it fail. Be opportunistic.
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -168,6 +168,11 @@ fi
 
 sed -i -e 's/DEBUG = False/DEBUG = True/' pulpcore/pulpcore/app/settings.py
 
+{% if not tasking_allow_async_unsafe -%}
+# Patch DJANGO_ALLOW_ASYNC_UNSAFE out of the pulpcore tasking_system
+# Don't let it fail. Be opportunistic.
+sed -i -e '/DJANGO_ALLOW_ASYNC_UNSAFE/d' ../pulpcore/pulpcore/tasking/entrypoint.py || true
```
fao89, line 174: shouldn't it be done at `pulpcore/content/__init__.py` also?

mdellweg: No, the content app is a different story...

dralley:
We should file an issue to revert this near the end of the 3.17 cycle, once we're ready to make it mandatory.

mdellweg:
> We should file an issue to revert this near the end of the 3.17 cycle, once we're ready to make it mandatory.

I was thinking about this. But maybe we should keep it around for long lived release branches.

ipanova (approved):

PR #502, opened 2021-09-21
Enable azure tests

### LINKED ISSUE #501: As plugin writer, I can test azure storage

fao89:
closes #501

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -30,6 +30,9 @@ jobs:
           {%- if docs_test %}
           - TEST: docs
           {%- endif %}
+          {%- if test_azure %}
+          - TEST: azure
+          {%- endif %}
           {%- if test_s3 %}
           - TEST: s3
           {%- endif %}
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -123,7 +123,7 @@ VARSYAML
 
 {%- if docker_fixtures %}
 
-if [[ "$TEST" == "pulp" || "$TEST" == "performance" || "$TEST" == "upgrade" || "$TEST" == "s3" || "$TEST" == "plugin-from-pypi" ]]; then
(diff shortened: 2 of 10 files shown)


templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -30,6 +30,9 @@ jobs:
           {%- if docs_test %}
           - TEST: docs
           {%- endif %}
+          {%- if test_azure %}
+          - TEST: azure
+          {%- endif %}
```
fao89, line 35: I've been thinking about only enabling it at nightly job

mdellweg: I would say, new tests should run more often. What they are supposed to test is probably a hot area of code change. Once you are confident about them, you can move them to nightly only.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -149,6 +149,25 @@ minio_secret_key: "'$MINIO_SECRET_KEY'"' vars/main.yaml
...
+
+if [ "$TEST" = "azure" ]; then
+  mkdir -p azurite
+  cd azurite
+  openssl req -newkey rsa:2048 -x509 -nodes -keyout azkey.pem -new -out azcert.pem -sha256 -days 365 -addext "subjectAltName=DNS:pulp-azurite" -subj "/C=CO/ST=ST/L=LO/O=OR/OU=OU/CN=CN"
+  sudo cp azcert.pem /usr/local/share/ca-certificates/azcert.crt
+  sudo dpkg-reconfigure ca-certificates
+  cd ..
+  sed -i -e '/^services:/a \
+  - name: pulp-azurite\
```
mdellweg, line 162: As if we build our own docker compose. ;)

mdellweg (commented):
Looks good so far. I would like to see it in action.

fao89:
> Looks good so far. I would like to see it in action.

https://github.com/pulp/pulp_ansible/pull/649

templates/github/.github/workflows/nightly.yml.j2
fao89: I was worried about the CI load, but I agree with you

mdellweg (approved):

PR #504, opened 2021-09-27
Use the base_ref when checking for single commit job

daviddavis:
Let's see if this fixes the issue.

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -20,7 +20,7 @@ jobs:
   {%- if single_commit_check %}
   single_commit:
     name: Assert single commit
-    if: github.ref == 'refs/heads/{{ plugin_default_branch }}'
+    if: github.base_ref == '{{ plugin_default_branch }}'
     steps:
     - name: Checkout
       uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -20,7 +20,7 @@ jobs:
   {%- if single_commit_check %}
   single_commit:
     name: Assert single commit
-    if: github.ref == 'refs/heads/{{ plugin_default_branch }}'
+    if: github.base_ref == '{{ plugin_default_branch }}'
```
mdellweg, line 23: Docs state this is only available if the trigger is a pull request. I am willing to assume that the condition nicely expands to `False` in all other cases.

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

mdellweg (approved):

templates/github/.github/workflows/ci.yml.j2
daviddavis: That was my assumption as well. I tested it out to confirm though:

https://github.com/daviddavis/pulpcore/commit/7343fb2ea5c9fdc90c01274a2d99008927fe80a9
https://github.com/daviddavis/pulpcore/actions/runs/1278196464

PR #521, opened 2021-11-03
Consider a runner with disabled redis services

### LINKED ISSUE #524: Enable plugin writers to disable the Redis service for specific test runners

lubosmj:
closes #524

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -106,29 +106,84 @@ jobs:
           echo ::group::HTTPIE
           pip install httpie
           echo ::endgroup::
+          echo "HTTPIE_CONFIG_DIR=$GITHUB_WORKSPACE/.ci/assets/httpie/" >> $GITHUB_ENV
+
+      - name: Set environment variables
+        run: |
           echo "TEST=${{ matrix.env.TEST }}" >> $GITHUB_ENV
 
       - name: Before Install
+
         run: .github/workflows/scripts/before_install.sh
         shell: bash
-
-      - name: Install
-        run: .github/workflows/scripts/install.sh
         env:
           PY_COLORS: '1'
           ANSIBLE_FORCE_COLOR: '1'
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
+
+      - name: Install
+
+        run: .github/workflows/scripts/install.sh
         shell: bash
+        env:
+          PY_COLORS: '1'
+          ANSIBLE_FORCE_COLOR: '1'
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
 
       - name: Install Python client
+
         run: .github/workflows/scripts/install_python_client.sh
+        shell: bash
+
+      - name: Install Ruby client
+        if: ${{ env.TEST == 'bindings' }}
+        run: .github/workflows/scripts/install_ruby_client.sh
+        shell: bash
 
       - name: Before Script
-        run: |
-          .github/workflows/scripts/before_script.sh
+
+        run: .github/workflows/scripts/before_script.sh
+        shell: bash
+        env:
+          PY_COLORS: '1'
+          ANSIBLE_FORCE_COLOR: '1'
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
+          REDIS_DISABLED: ${{ contains('s3', matrix.env.TEST) }}
+
+      - name: Setting secrets
+        if: github.event_name != 'pull_request'
+        run: python3 .github/workflows/scripts/secrets.py "$SECRETS_CONTEXT"
+        env:
+          SECRETS_CONTEXT: ${{ toJson(secrets) }}
 
       - name: Script
+
         run: .github/workflows/scripts/script.sh
         shell: bash
+        env:
+          PY_COLORS: '1'
+          ANSIBLE_FORCE_COLOR: '1'
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


lubosmj:
This commit might be useful across all the plugins since we probably want to test the pulp services without the Redis server. Modifying the `script.sh` is not recommended by default (https://github.com/pulp/pulpcore/pull/1663/files#diff-02059c243ab1ae5faa6a95ec965f0b351e0f6c96e5452c90881b6b7e3ee11618). At the moment, this change enables setting up the configuration for the fixtures.

I am thinking about introducing a new option to the template config. Not sure if it will be that needed, though.

mdellweg:
I would prefer, if we could fold this into on of s3 or azure tests. It should be reasonably independent to not need another full excercise of the whole test suite.

dralley:
I agree with @mdellweg

lubosmj:
I have introduced a new configuration option. Please take a look at the changes, @mdellweg.

.github/workflows/bootstrap.yaml
```diff
@@ -72,6 +72,8 @@ jobs:
     runs-on: ubuntu-latest
     # run only after lint finishes
     needs: lint
+    env:
+      REDIS_DISABLED: "azure s3"
```
dralley, line 76: Don't disable caching for both, we still want some coverage of a non-local storage backend in combination with the cache.

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -91,6 +91,12 @@ jobs:
     runs-on: ubuntu-latest
     # run only after lint finishes
     needs: lint
+    env:
+    {%- if disabled_redis_runners %}
+      REDIS_DISABLED: "{{ disabled_redis_runners | join(' ') }}"
+    {%- else %}
+      REDIS_DISABLED: ""
```
mdellweg, line 98: I'd rather have a `REDIS_DISABLED=true/false` a few lines down in the matrix.

Some thing like maybe:
`REDIS_DISABLED: "{{ "s3" in (disabled_redis_runners | default([])) }}"`

I think, this makes the generated workflows much more readable. Maybe it's also possible outside of the matrix with
```
env:
  - REDIS_DISABLED: ${{ matrix.env.TEST == "s3" }}
```
This way the workflows would keep their current names.

templates/github/.github/workflows/scripts/before_script.sh.j2
```diff
@@ -29,6 +29,14 @@ if [[ "$TEST" == 'pulp' || "$TEST" == 'performance' || "$TEST" == 'upgrade' || "
   cmd_prefix dnf install -yq lsof which dnf-plugins-core
 fi
 
+{%- if disabled_redis_runners %}
+read -ra REDIS_DISABLED <<< "$REDIS_DISABLED"
+if [[ " ${REDIS_DISABLED[*]} " =~ " ${TEST} " ]]; then
+  cmd_prefix bash -c "s6-svc -d /var/run/s6/services/redis"
+  echo "The Redis service was disabled for $TEST"
+fi
+{%- endif %}
```
mdellweg, line 38: Do we have a way to disable Redis also? Or is Pulp just able to fallback if missing?

lubosmj: This disables the Redis server service. What do you mean by `disable Redis also`? The idea, that comes from https://pulp.plan.io/issues/8997, was to simulate the Pulp's behaviour after disabling the Redis service. The actual change to the template allows plugin writers to test their plugins in such a state as well.

templates/github/.github/workflows/ci.yml.j2
lubosmj: I do not know where else a few lines down in the matrix you could define an environment variable. I think I cannot do this:
```
    strategy:
      fail-fast: false
      matrix:
        env:
          - TEST: pulp
          - TEST: docs
          - TEST: azure
            - REDIS_DISABLED: false
          - TEST: s3
            - REDIS_DISABLED: true
```

mdellweg: you could do:
```
    strategy:
      fail-fast: false
      matrix:
        env:
          - TEST: pulp
          - TEST: docs
          - TEST: azure
            REDIS_DISABLED: false
          - TEST: s3
            REDIS_DISABLED: true
```
But on a second thought, this changes the names of the workflow jobs and that'd need adjusting the repo settings everywhere.
So i'd really go with an `env` stanza in the `steps:` part of the workflow.

templates/github/.github/workflows/scripts/before_script.sh.j2
mdellweg: I'd say this shuts the service down, simulating a failure. But what if a user wants to run Pulp without ever installing / configuring redis. Would this use different codepaths in any way? Do we not support such a thing yet?

lubosmj: There is a separate issue filed for that: https://pulp.plan.io/issues/9475. Should I rather focus on that one in this PR instead?

mdellweg: No, just wanted to know if we miss an opportunity today.

lubosmj:
@mdellweg, I made the changes you have proposed. Now, the environment variable `REDIS_DISABLED` is defined in the place where all other environment variables reside for each runner.

plugin-template
```diff
@@ -66,6 +66,7 @@ DEFAULT_SETTINGS = {
     "test_cli": False,
     "test_performance": False,
     "test_released_plugin_with_next_pulpcore_release": False,
+    "disabled_redis_runners": [],
```
mdellweg, line 69: Wouldn't it be the easiest to just default to "s3" here? It shouldn't fail if "s3" wasn't even tested, no?

templates/github/.github/workflows/scripts/before_script.sh.j2
```diff
@@ -29,6 +29,11 @@ if [[ "$TEST" == 'pulp' || "$TEST" == 'performance' || "$TEST" == 'upgrade' || "
   cmd_prefix dnf install -yq lsof which dnf-plugins-core
 fi
 
+if [[ "$REDIS_DISABLED" == true ]]; then
+  cmd_prefix bash -c "s6-svc -d /var/run/s6/services/redis"
+  echo "The Redis service was disabled for $TEST"
+fi
```
mdellweg, line 35: This is nice and readable!

.github/workflows/bootstrap.yaml
```diff
@@ -123,8 +123,10 @@ jobs:
         run: .github/workflows/scripts/install_python_client.sh
 
       - name: Before Script
-        run: |
-          .github/workflows/scripts/before_script.sh
+        run: .github/workflows/scripts/before_script.sh
+        shell: bash
+        env:
+          REDIS_DISABLED: ${{ contains('s3', matrix.env.TEST) }}
```
mdellweg, line 129: I sense that this is not exactly the same as in the file that gets generated. (Not your fault!) Can you regenerate it please?
@fao do you see a way that we can asure this part of the file equals the on the fly generated bit by bit?

templates/macros.j2
```diff
@@ -81,14 +86,14 @@ GITHUB_CONTEXT: {{ "${{ github.event.pull_request.commits_url }}" }}
 {%- endmacro -%}
 
 
-{%- macro run_script(name, file, withenv=True, condition=None) -%}
+{%- macro run_script(name, file, withenv=True, condition=None, customenvs=None) -%}
```
mdellweg, line 89: I would name that variable `extra_env` or just `env`.

plugin-template
lubosmj: It should not fail. Yet, I think that we should not disable Redis for any storage by default.

.github/workflows/bootstrap.yaml
fao89: I saw this tweet yesterday:
https://twitter.com/github/status/1461106904122417156?s=20
about reusable workflows, I didn't read the article, but maybe this is the solution

lubosmj: No, the question was about comparing two files on the fly. We are not going to run another workflow right now.

templates/github/.github/workflows/scripts/before_script.sh.j2
```diff
@@ -29,6 +29,11 @@ if [[ "$TEST" == 'pulp' || "$TEST" == 'performance' || "$TEST" == 'upgrade' || "
   cmd_prefix dnf install -yq lsof which dnf-plugins-core
 fi
 
+if [[ "$REDIS_DISABLED" == true ]]; then
```
mdellweg, line 32: This fails here:
https://github.com/pulp/pulpcore/runs/4319788337?check_suite_focus=true#step:12:143

```suggestion
if [[ "${REDIS_DISABLED:-false}" == true ]]; then
```

mdellweg (approved):

PR #522, opened 2021-11-03
Publish the changelog from default branch

### LINKED ISSUE #523: Changelogs for bugfix releases are not immediately published
This is a transformation of https://pulp.plan.io/issues/9361

mdellweg:
This will automatically publish the changelog to the docs root directory
whenever a changelog cherry pick is merged.

fixes #523

### DIFF
--- a/templates/github/.github/workflows/changelog.yml.j2
+++ b/templates/github/.github/workflows/changelog.yml.j2
@@ -1,0 +1,45 @@
+{% include 'header.j2' %}
+{% from 'macros.j2' import
+  checkout,
+  setup_python,
+  set_secrets,
+  install_python_deps,
+with context %}
+---
+name: {{ plugin_camel_short | default("Pulp") }} changelog update
+on:
+  push:
+    branches:
+      - {{ plugin_default_branch }}
+    paths:
+      - CHANGES.rst
+      - CHANGES.md
+
+jobs:
+
+  update-changelog:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    steps:
+      {{ checkout() | indent(6) }}
+
+      {{ setup_python() | indent(6) }}
+
+      {{ install_python_deps(deps="-r doc_requirements.txt") | indent(6) }}
+
+      - name: Fake api schema
+        run: |
+          mkdir -p docs/_build/html
+          echo "{}" > docs/_build/html/api.json
+          mkdir -p docs/_static
+          echo "{}" > docs/_static/api.json
+      - name:
+        run: make html
+        working-directory: ./docs
+        env:
+          PULP_CONTENT_ORIGIN: "http://localhost/"
+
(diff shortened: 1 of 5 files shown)


mdellweg:
I don't yet know, how to test this.

mdellweg:
The building part seems to work now.

bmbouter:
@dkliban can you remind me of the layout of the docs on pulpproject.org ? I'm wondering specifically if we have "nightly" for each version of pulpcore for example.

bmbouter:
@mdellweg is the goal of this PR to push docs with every merge (pre-release right?) I was thinking yes because the release automation handles the pushing in other cases already (is that right?)

mdellweg:
This PR intends to target the auto-generated "cherry-pick changelog" PR's specifically. On merge, they change the `CHANGES.rst` on the main branch, and that is the trigger for the new workflow. It does not touch any of the docs in any sub-directory, but only updates the most prominent place of the changelog.

bmbouter:
I see, so the idea is when a say 3.15.z changelog is merged to master, also push it to docs.pulpproject.org/latest/ is that right?

If my understanding is right, personally I don't think the solution and maintaining it is worth the benefit. The changelog will be right the next time a release happens, and this is getting complex. That's my take at least.

mdellweg:
> The changelog will be right the next time a release happens

That is exactly the problem that was reported multiple times.

bmbouter:
> > The changelog will be right the next time a release happens
> 
> That is exactly the problem that was reported multiple times.

Yup and they're right, but to me it's not worth the complexity. If others want to review it and go forward with this, I'm ok with that.

templates/github/.github/workflows/changelog.yml.j2
```diff
@@ -0,0 +1,45 @@
...
+with context %}
+---
+name: {{ plugin_camel_short | default("Pulp") }} changelog update
+on:
+  push:
+    branches:
+      - {{ plugin_default_branch }}
+    paths:
+      - CHANGES.rst
+      - CHANGES.md
```
fao89, line 16: I think it should be:
```suggestion
      - CHANGES/**
```

mdellweg: I don't think so. We want to republish the generated changelog (either rst or md) and not every time something happened with the "unreleased" snippet files.

fao89: Oh! I misunderstood the PR, I thought it was for building changelogs whenever a new entry was added.

mdellweg: No, it's publishing the complete changelog whenever it was changed.

fao89 (approved):
approved but it needs a rebase

fao89 (approved):
LGTM

dkliban (approved):

PR #530, opened 2021-12-06
Set an upper bound for mistune

lubosmj:
This temporary change is required to resolve issues in the CI pipeline (docs runner).

The change is inspired by https://github.com/miyakogi/m2r/issues/66.

[noissue]


mdellweg:
Somehow it doesn't seem to resolve the issue.

lubosmj:
@mdellweg, obviously, because of the pulpcore's PR that should be merge first. Is that correct?

mdellweg (approved):

PR #531, opened 2021-12-09
Update every plugin CI job

fao89:
we can trigger the update job for every plugin when we have a big CI change

### DIFF
--- a/.github/workflows/update_ci.yml
+++ b/.github/workflows/update_ci.yml
@@ -1,0 +1,36 @@
+---
+name: CI Update
+on:
+  workflow_dispatch:
+
+jobs:
+  update:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_ci.yml
```diff
@@ -0,0 +1,42 @@
+# WARNING: DO NOT EDIT!
+#
+# This file was generated by plugin_template, and is managed by it. Please use
+# './plugin-template --github pulp_ansible' to update this file.
+#
+# For more info visit https://github.com/pulp/plugin_template
```
mdellweg, line 6: This does not apply, right?

mdellweg (approved):
I want to see it in action. ;)

PR #534, opened 2021-12-14
Update references from master to main.

ipanova:
[noissue]


fao89 (approved):
Thank you!

mdellweg (approved):
I think this should come with a rename in this repo.

PR #539, opened 2021-12-22
Fix x.y docs publishing

fao89:
[noissue]


goosemania:
Looks good!
With this, docs should be up to date for /x.y./ docs.
@dkliban , fyi, in case we intentionally skipped x.y. docs and published only x.y.z.

goosemania (approved):

PR #542, opened 2022-01-06
Add workflow_dispatch to nightly workflow

fao89:
[noissue]


bmbouter:
If this does what I think it does, +1!

fao89:
> If this does what I think it does, +1!

it enables us to trigger the nightly job when we want to see how the main branch is

mdellweg (approved):

PR #545, opened 2022-01-14
Add a CI job that collects on deprecations

mdellweg:
This will appear as a separate job in the workflow that can fail
individually and be marked non essential.

[noissue]


mdellweg:
See how it works here:

https://github.com/pulp/pulp_npm/pull/141

fao89 (approved):

PR #547, opened 2022-01-14
Add plugin template parallel running supports

### LINKED ISSUE #546: Functional tests marked with `parallel` are run in parallel
To allow for quicker test running in the CI, it would be good to split the single call to the functional tests into two calls. The first should collect all tests marked with `parallel` and run those using the `-n` param provided by `pytest-xdist`. After that it should collect all tests not marked with `parallel` and run those just like it does today.

bmbouter:
Adds the `parallel_test_workers` which defaults to 8. This runs any test 
marked with `@pytest.mark.parallel` to run in parallel with N concurrent
processes. Setting this value to 0 disables parallel running altogether.

closes #546

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -37,6 +37,7 @@ DEFAULT_SETTINGS = {
     "docker_fixtures": False,
     "docs_test": True,
     "issue_tracker": "github",
+    "parallel_test_workers": 8,
     "pulp_scheme": "https",
     "plugin_app_label": None,
     "plugin_camel": None,
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -186,7 +186,12 @@ fi
 if [ -f $FUNC_TEST_SCRIPT ]; then
   source $FUNC_TEST_SCRIPT
 else
+{% if parallel_test_workers == 0 %}
     pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
+{% else %}
+    pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs {{ plugin_snake }}.tests.functional -m parallel -n {{ parallel_test_workers }}
+    pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not parallel"
+{% endif %}
 fi
 
 {%- if test_cli %}
--- a/CHANGES/546.feature
+++ b/CHANGES/546.feature
@@ -1,0 +1,4 @@
+Added the ``parallel_test_workers`` template_config.yml variable which defaults to 8. This variable
+specifies the number of processes to run tests in parallel with. Tests can be marked as runnable in
+parallel with the ``@pytest.mark.parallel`` decorator. Setting this variable to 0 disables parallel
+running entirely.
--- a/README.md
+++ b/README.md
@@ -144,10 +144,15 @@ The following settings are stored in `template_config.yml`.
   aiohttp_fixtures_origin
                         Identifies the network origin of the aiohttp fixtures.
 
-  issue_tracker         Which issue tracker the project will use. Valid values are 'redmine' and 'github'.
-                        To switch from Redmine to GitHub use the --migrate-github-issues option.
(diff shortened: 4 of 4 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -186,7 +186,12 @@ fi
 if [ -f $FUNC_TEST_SCRIPT ]; then
   source $FUNC_TEST_SCRIPT
 else
+    {% if parallel_test_workers == 0 %}
     pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
+    {% else -%}
+    pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m parallel -n {{ parallel_test_workers }}
```
mdellweg, line 192: you need to configure that marker for pytest to avoid warnings

plugin-template
```diff
@@ -37,6 +37,7 @@ DEFAULT_SETTINGS = {
     "docker_fixtures": False,
     "docs_test": True,
     "issue_tracker": "github",
+    "parallel_test_workers": 0,
```
mdellweg, line 40: I just started to ask myself, whether it is worth making this an option. Current tests will not be accidentally marked as parallel.
The value i see in this option is if you intend to make parallel the default and want plugins do declare tests a serial.

mdellweg: Oh, i just realized, it's not-a-boolean for a reason.

fao89 (commented):
you need to add a changelog entry

templates/github/.github/workflows/scripts/script.sh.j2
bmbouter: I agree, but I don't think the plugin template can append that config to pyproject.toml can it? It would be great if it could.

plugin-template
bmbouter: I like the idea that the number of parallelization is configurable, but I do think it would be better if we could make the default 8 instead of zero. The one blocking reason is we need every plugin to receive `pytest-xdist`. Right now I have pulpcore enabling that [here](https://github.com/pulp/pulpcore/pull/1810/files#diff-93394e9ecb9acc4a48dffa8515bf0d6c45d2170dd7fd2dcc76d1bd722d94fa0fR8) to do that. If we wanted to make this 8 we could have pulp-smash declare `pytest-xdist` as a dependency.... I think I should do that, merge it, and then change this default to 8. What do you think about this proposal?

mdellweg: Sounds like a plan. My underlying question is still, in the long run do you expect we mark tests as viable for parallel or mark them for needsbe serial?

bmbouter: In the very long run I expect the mark to be serial. To do that though I'd have to have all tests auto-mark as `serial` and to do that I think projects would have to take an action that I can't provide automatically to them in the plugin template. So for that reason I think leaving it as parallel for now is the best I can do. Wdyt? Suggestions are welcome.

bmbouter: Also this is ready: https://github.com/pulp/pulp-smash/pull/1266

templates/github/.github/workflows/scripts/script.sh.j2
bmbouter: Actually I was able to do this here:  https://github.com/pulp/pulp-smash/pull/1268/files

bmbouter:
I changed the default to 8 and added the changelog.

bmbouter:
Here's the parallel part in action!! https://github.com/pulp/pulpcore/runs/4845911293?check_suite_focus=true#step:13:662

templates/github/.github/workflows/scripts/script.sh.j2
mdellweg: At least for new bootstraps it should. Also this is kind of a reference to update those files.

And in general, yes there is room for improvement. e.g. it's a pain to update the docs machinery from the template. The CI parts is really the best maintained part.

mdellweg: I see them moved to pulp-smash.

mdellweg (approved):

PR #550, opened 2022-01-18
Always display deprecation warnings

mdellweg:
[noissue]


.github/workflows/bootstrap.yaml
```diff
@@ -208,6 +208,7 @@ jobs:
 
   deprecations:
     runs-on: ubuntu-latest
+    if: always()
```
fao89, line 211: 👍🏽

fao89 (approved):

PR #558, opened 2022-01-26
WIP: Use default flake8 config name

quba42:
Renaming `flake8.cfg` to `.flake8` eliminates the need to run `flake8` with the option `--config flake8.cfg`.


plugin-template
```diff
@@ -98,6 +98,9 @@ DEPRECATED_FILES = {
         "docs/_static/.this_makes_it_so_git_keeps_the__static_folder",
         "docs/_static/api.json",
     ],
+    "bootstrap": [
+        "flake8.cfg",
+    ],
```
quba42, line 103: Major problem: flake8.cfg was seeded as part of bootstrap, so adding this change will break a plugins CI unless flake8.cfg is renamed to .flake8, but that won't be warned for by `./plugin-template --github`...

mdellweg (commented):
Looks like i started a review and never submitted it...
I like the idea. Also it's worth thinking about just moving the flake config to the github section. We allow the plugin writer to opt out using flake. So maybe we can be opinionated about the errors we ignore throughout the pulp project.
One thing this pr needs is that the old config be added to the deprecated files.

quba42:
@mdellweg, @bmbouter, @dkliban While I am happy to see the interest in this change, I don't think I will find the time to figure out how to force the rename of the file during plugin_template application any time soon. Feel free to adopt this PR (or close it and open a new one) if any of you do have the time. :wink:

mdellweg:
See #668

quba42:
has been implemented and merged on #668

PR #559, opened 2022-01-27
Run proxy.py before functional tests

fao89:
[noissue]


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -183,6 +183,10 @@ if [[ "$TEST" == "performance" ]]; then
   exit
 fi
 
+# Run proxy.py
+podman run -d -it -p 8899:8899 --name ciproxy --rm ghcr.io/abhinavsingh/proxy.py:latest --basic-auth foo:bar --hostname 0.0.0.0
```
mdellweg, line 187: can you add that container in the same way as we start the minio container?

mdellweg: I can see this being overengineered, but i'd like to see the CI containers to be started in one place.

mdellweg (approved):
Just two (kind of related) questions left:
Did you want to make this dependent on a template_config parameter?
What's the purpose of the `PULP_PROXY_TEST=enabled` variable? If tests are supposed to react to it, i would expect to see a pulp-smash setting for it.

fao89:
> Just two (kind of related) questions left: Did you want to make this dependent on a template_config parameter? What's the purpose of the `PULP_PROXY_TEST=enabled` variable? If tests are supposed to react to it, i would expect to see a pulp-smash setting for it.

when proxy image isn't up and you run your tests, it would fail, I came up with this var to skip tests when it is not set.
A pulp-smash setting sounds good, but I don't know how to set up it for specific tests or have pulp-smash to run proxy image before starting the tests

mdellweg:
> > Just two (kind of related) questions left: Did you want to make this dependent on a template_config parameter? What's the purpose of the `PULP_PROXY_TEST=enabled` variable? If tests are supposed to react to it, i would expect to see a pulp-smash setting for it.
> 
> when proxy image isn't up and you run your tests, it would fail, I came up with this var to skip tests when it is not set. A pulp-smash setting sounds good, but I don't know how to set up it for specific tests or have pulp-smash to run proxy image before starting the tests

The idea behind pulp-smash is that it should help you to set up (and configure to your tests) a proxy setup both in CI and vagrant.
But we should ask @bmbouter if there are plans to provide a proxy as a pytest fixture.

bmbouter:
Yeah I was just commenting that we should do this differently. https://github.com/pulp/pulp_ansible/pull/802/files#diff-0b9d42dda1184c2f2c2da7e27b0fdc97a60a51063093fc7f88356e9910d9ca9fR206

We need to do it that way otherwise the tests are becoming tightly coupled to the CI environment which we need to move away from to enable others to run all the tests (local devs, downstreams, etc). Let's not enable this.

@fao89 should I or should you write this pytest fixture in pulp-smash?

bmbouter (changes requested):
I think we need to add this as a pytest fixture in pulp-smash instead, otherwise the tests that use this will become tightly coupled with this new requirement.

PR #564, opened 2022-02-01
Adds issue template

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/ISSUE_TEMPLATE/feature_request.md.j2
+++ b/templates/github/.github/ISSUE_TEMPLATE/feature_request.md.j2
@@ -1,0 +1,20 @@
+---
+name: ✨ Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature, Triage-Needed
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context about the feature request here.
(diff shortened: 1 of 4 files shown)


templates/github/.github/ISSUE_TEMPLATE/task.md.j2
```diff
@@ -0,0 +1,10 @@
+---
+name: 🗒️ Task
+about: Documentation changes, CI/CD changes, work items such as "investigate X"
+title: ''
+labels: Task, Triage-Needed
```
lubosmj, line 5: A task does not need to be flagged as `Triage-Needed`.

lubosmj (changes requested):
Would you mind also adding a template for stories?

templates/github/.github/ISSUE_TEMPLATE/feature_request.md.j2
```diff
@@ -0,0 +1,20 @@
+---
+name: ✨ Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature, Triage-Needed
```
lubosmj, line 5: A feature request should not be marked as `Triage-Needed` as well. Am I correct?

templates/github/.github/ISSUE_TEMPLATE/task.md.j2
dralley: Per the old process, yes.

But I don't think there's a great reason to exclude stories or tasks from the triage process personally.  They all ought to have some level of review.

fao89:
> Would you mind also adding a template for stories?

feature == stories

templates/github/.github/ISSUE_TEMPLATE/task.md.j2
fao89: on the previous process, 
the default was an issue and everything went to triage, and users rarely changed that.

dralley (approved):

lubosmj (approved):

PR #565, opened 2022-02-02
Improving issue template

fao89:
[noissue]


ipanova (approved):
looks good to me!

dralley (approved):

mdellweg:
Maybe it would be wise if we could switch off the need for the Triage-Needed label on a per plugin basis.

PR #568, opened 2022-02-04
Run nightly tests on dispatch

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -187,13 +187,13 @@ if [ -f $FUNC_TEST_SCRIPT ]; then
   source $FUNC_TEST_SCRIPT
 else
 {% if parallel_test_workers == 0 %}
-    if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
+    if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
     else
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not nightly"
     fi
 {% else %}
-    if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
+    if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs {{ plugin_snake }}.tests.functional -m parallel -n {{ parallel_test_workers }}
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not parallel"
     else


fao89:
Tested here: https://github.com/pulp/pulp_ansible/pull/832

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -187,13 +187,13 @@ if [ -f $FUNC_TEST_SCRIPT ]; then
...
 {% if parallel_test_workers == 0 %}
-    if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
+    if [[ "$GITHUB_EVENT_NAME" == "schedule" || "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]]; then
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
     else
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not nightly"
     fi
 {% else %}
-    if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
+    if [[ "$GITHUB_EVENT_NAME" == "schedule" || "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]]; then
```
mdellweg, line 196: Maybe I'm amiss here. But we only want to do this on nighty runs. Should we filter by the workflow too?

fao89: @bmbouter wanted to run it on dispatch, so I started this PR,
I don't have a strong opinion on it

mdellweg: Oh, i thought we would have a way to "dispatch a nightly run by hand".

bmbouter: Yeah when I went to run the "nightly" by hand @fao89 pointed out that is a different dispatch type. So he kindly made this for me which would run the "nightly" behavior both when dispatched by GHI or by a human running it manually.

mdellweg: What i want to point at is that we should try not to guess whether it's a "nightly" from the way the job was triggered, but try to be dependent, on whether the nightly workflow was triggered.
When i rerun a PR CI i probably do not want to trigger this.

bmbouter: We never want the nightly tests to run at PR time, so if this PR is causing that to change we should do this differently. I don't know enough about GHI to know.

mdellweg: It still sounds like asking for the presence of strawberries, because you want apples...

bmbouter: I don't know about apples or strawberries, but I do want the same tests run in that job regardless of how its dispatched.

mdellweg: And with "that job" you mean any job triggered by this workflow file: https://github.com/pulp/plugin_template/blob/main/templates/github/.github/workflows/nightly.yml.j2
And that is what the "if" condition is meant to determine, but it asks for strange clues to get there, no?
I believe it will work today, because no other workflow that has a dispatch trigger will call "script.sh". But will that stay true?

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -187,13 +187,13 @@ if [ -f $FUNC_TEST_SCRIPT ]; then
   source $FUNC_TEST_SCRIPT
 else
 {% if parallel_test_workers == 0 %}
-    if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
+    if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
```
mdellweg, line 190: This sounds _way_ more specific to me.

mdellweg (approved):

PR #569, opened 2022-02-15
Removes use of new-pulpcore-worker and resource-manager

dkliban:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -134,12 +134,8 @@ if [[ "$TEST" == "upgrade" ]]; then
   # Restarting single container services
   cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-api"
   cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-content"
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -134,12 +134,8 @@ if [[ "$TEST" == "upgrade" ]]; then
   # Restarting single container services
   cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-api"
   cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-content"
-  cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-resource-manager"
   cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-worker@1"
   cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-worker@2"
```
fao89, line 138: we don't want to stop them, so I guess it is -r now?
```suggestion
  cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-worker@1"
  cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-worker@2"
```

fao89 (commented):
Actually, this is the update scenario, I believe the container image for this scenario should be able to run both tasking systems

PR #570, opened 2022-02-16
Use upgrade image on upgrade tests

fao89:
[noissue]


fao89:
Tested here: https://github.com/pulp/pulpcore/pull/2227

ipanova (approved):

PR #574, opened 2022-02-17
Ensure changelog entries are removed on y stream releases

fao89:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/stage-changelog-for-default-branch.py.j2
+++ b/templates/github/.github/workflows/scripts/stage-changelog-for-default-branch.py.j2
@@ -61,9 +61,10 @@ except GitCommandError:
         git.cherry_pick("--continue")
 git.reset("origin/{{ plugin_default_branch }}")
 
-# Do not remove changelog entries
-msg = repo.commit().message
-git.reset("HEAD~1")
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/scripts/stage-changelog-for-default-branch.py.j2
```diff
@@ -61,9 +61,10 @@ except GitCommandError:
         git.cherry_pick("--continue")
 git.reset("origin/{{ plugin_default_branch }}")
 
-# Do not remove changelog entries
-msg = repo.commit().message
-git.reset("HEAD~1")
-git.add("CHANGES.rst")
-git.commit("-m", msg)
-git.stash()
+if not release_version_arg.endswith(".0"):
```
mdellweg, line 64: So this will make the changelog snippets from minor releases (3.17.0) disappear from main, but not the backported fix from e.g. 3.17.1?

mdellweg (approved):

templates/github/.github/workflows/scripts/stage-changelog-for-default-branch.py.j2
fao89: yes

PR #579, opened 2022-02-24
Test pulp in rerouted environment

### LINKED ISSUE #578: Add tests to run pulp in a rerouted environment
For now, we will just reroute the container in the s3 test environment.

mdellweg:
fixes #578

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -150,6 +150,9 @@ if [ "$TEST" = "s3" ]; then
   sed -i -e '$a s3_test: true\
 minio_access_key: "'$MINIO_ACCESS_KEY'"\
 minio_secret_key: "'$MINIO_SECRET_KEY'"' vars/main.yaml
+
+  echo "PULP_API_ROOT=/rerouted/djnd/" >> "$GITHUB_ENV"
+  export PULP_API_ROOT="/rerouted/djnd/"
 fi
 {%- endif %}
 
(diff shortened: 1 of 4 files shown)


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -150,6 +150,9 @@ if [ "$TEST" = "s3" ]; then
   sed -i -e '$a s3_test: true\
 minio_access_key: "'$MINIO_ACCESS_KEY'"\
 minio_secret_key: "'$MINIO_SECRET_KEY'"' vars/main.yaml
+
+  echo "PULP_API_ROOT=/rerouted/djnd/" >> "$GITHUB_ENV"
+  export PULP_API_ROOT="/rerouted/djnd/"
```
mdellweg, line 155: This should really be the lift of that feature. Do we want to set the vars in a different way?

cc @fao89

mdellweg:
Results here:

https://github.com/pulp/pulp_file/pull/644

templates/github/.github/workflows/scripts/install.sh.j2
fao89: I think this is good

fao89 (approved):

mdellweg:
How do we want to handle failing tests once this gets rolled out to plugins?

fao89:
> How do we want to handle failing tests once this gets rolled out to plugins?

I would trigger the update job and fix the tests on plugins that break

mdellweg:
> > How do we want to handle failing tests once this gets rolled out to plugins?
> 
> I would trigger the update job and fix the tests on plugins that break

I'd take at least core and file. (Have them already.)

With that i think we are good to go.

PR #582, opened 2022-03-01
Option to disable reroute tests

fao89:
Updating pulpcore branches may break with reroute
https://github.com/pulp/pulpcore/runs/5379632837?check_suite_focus=true#step:8:511

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -68,6 +68,7 @@ DEFAULT_SETTINGS = {
     "test_deprecations": True,
     "test_performance": False,
     "test_released_plugin_with_next_pulpcore_release": False,
+    "test_reroute": True,
     "disabled_redis_runners": [],
     "test_azure": False,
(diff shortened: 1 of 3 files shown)


plugin-template
```diff
@@ -68,6 +68,7 @@ DEFAULT_SETTINGS = {
     "test_deprecations": True,
     "test_performance": False,
     "test_released_plugin_with_next_pulpcore_release": False,
+    "test_reroute": True,
```
bmbouter, line 71: What was the motivation for making it optional?

fao89: https://github.com/pulp/pulpcore/runs/5379632837?check_suite_focus=true#step:8:511

bmbouter: Make sense, thanks!

bmbouter (approved):

PR #583, opened 2022-03-02
Remove django_allow_async_unsafe handling

dralley:
[noissue]


dralley:
Caveat, 3.14 still needs this IIRC

dralley:
Removing the label because we can't merge this yet but we still need to do it eventually.

dralley:
We can probably do this now if we were to stop updating the 3.14 CI

PR #584, opened 2022-03-02
Fix invalid yaml syntax

bmclaughlin:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -72,7 +72,7 @@ jobs:
       {%- endif %}
 
       - name: Run extra lint checks
-        run: [ -x .ci/scripts/extra_linting.sh ] && .ci/scripts/extra_linting.sh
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -71,8 +71,10 @@ jobs:
         run: flake8 --config flake8.cfg
       {%- endif %}
 
+      {% if extra_linting -%}
       - name: Run extra lint checks
-        run: [ -x .ci/scripts/extra_linting.sh ] && .ci/scripts/extra_linting.sh
```
mdellweg, line 75: Would you mind if just quoting properly would have solved the problem?
maybe: `"run: [ -x .ci/scripts/extra_linting.sh ] && .ci/scripts/extra_linting.sh"`

mdellweg: Sorry, i did the wrong quotes in the first place.

mdellweg (approved):

PR #587, opened 2022-03-10
Update changelog history

fao89:
[noissue]


fao89:
ref: https://github.com/pulp/pulpcore/pull/2300

templates/github/.ci/scripts/changelog.py.j2
```diff
@@ -0,0 +1,75 @@
...
+    ).text
+
+
+def get_changelog_releases(changelog):
+    """
+    Get changelog entries for all releases
+
+    """
+    versions = re.findall(r"([0-9]+)\.([0-9]+)\.([0-9]+) \(", changelog)
+    return {".".join(v) for v in versions}
```
mdellweg, line 28:
```suggestion
    Get all versions in changelog.

    """
    versions = re.findall(r"([0-9]+\.[0-9]+\.[0-9]+) \(", changelog)
    return {v[0] for v in versions}
```
Assuming version behaves like a list of lists.

fao89: it returns  a list [X, Y, Z]
```
> /home/faguiard/project/pulpcore/.ci/scripts/changelog.py(29)get_changelog_releases()
-> return {".".join(v) for v in versions}
(Pdb) versions [0]
('3', '18', '1')
(Pdb) ".".join(versions[0])
'3.18.1'
(Pdb) 
```

dkliban (approved):

templates/github/.ci/scripts/changelog.py.j2
mdellweg: My point is that we don't need 3 capture groups for the individual numbers if we could also use one capture group for the version in its entirety.

PR #588, opened 2022-03-10
Fix azure tests

fao89:
[noissue]


ipanova (approved):

ipanova:
hm..pulpcore ci still seems to fail :/

PR #591, opened 2022-03-11
Add gitleaks to template

fao89:
[noissue]


templates/bootstrap/.gitleaks.toml.j2
```diff
@@ -0,0 +1,5 @@
+[allowlist]
+  description = "Our test install exports a test only MINIO ACCESS KEY"
+  paths = [
+    ".github/workflows/scripts/install.sh",
```
mdellweg, line 4: This is quite a complex file. I think we should extract all the key material into a dedicated keyfile and have all the scripts scanned in the regular fashion.

PR #593, opened 2022-03-16
Keep CI update on the latest 5 branches

fao89:
[noissue]


fao89:
tested here https://github.com/pulp/pulp_npm/pull/162

quba42:
Can we make it configurable in the plugin template how many release branches CI update should be enabled for?

README.md
```diff
@@ -148,6 +148,9 @@ The following settings are stored in `template_config.yml`.
                         'github'. To switch from Redmine to GitHub use the --migrate-github-issues
                         option.
 
+  keep_ci_update_for_latest_branches
+                        Total of branches for which CI updates are kept. The default is 5.
```
quba42, line 152: Just since I am not 100% sure based on reading the code: `5` here means: 5 release branches of the format `[[:digit:]]\.[[:digit:]]*` plus the default branch (`main`), so six branches total?

quba42: Actually, I can't see where the default branch is handled in this change, so I am guessing it is simply separate.

fao89: yes, it would be 5 x.y + main,
the main branch is where the x.y branches are triggered (so it is already running on main)

templates/github/.github/workflows/update_ci.yml.j2
```diff
@@ -20,6 +20,28 @@ jobs:
...
       - uses: actions/checkout@v2
 
+      - name: Set short_ref
+        id: vars
+        run: echo ::set-output name=short_ref::${GITHUB_REF#refs/*/}
+
+      - name: Dispatching
+        if: github.event_name == 'schedule'
+        run: |
+          for version in $(git ls-remote --heads https://github.com/pulp/{{ plugin_name }}.git | grep -o "[[:digit:]]\.[[:digit:]]*" | sort -V | tail -{{ keep_ci_update_for_latest_branches }})
```
mdellweg, line 30: This will not work with plugins living outside of the pulp org.

plugin-template
```diff
@@ -36,7 +36,9 @@ DEFAULT_SETTINGS = {
     "deploy_to_pypi": True,
     "docker_fixtures": False,
     "docs_test": True,
+    "github_org": "pulp",
```
mdellweg, line 39: Looking at you @pulp/galaxy

mdellweg (approved):
LGTM

PR #601, opened 2022-03-25
Pin jinja for docs publication

fao89:
[noissue]


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -17,6 +17,7 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   pip install -r ../pulpcore/doc_requirements.txt
 {%- endif %}
   pip install -r doc_requirements.txt
+  pip install "Jinja2<3.1"
```
mdellweg, line 20: Should this go into doc_requirements?

fao89: https://github.com/pulp/pulpcore/pull/2429

mdellweg (approved):
Right. There is no requirements file for this...

PR #602, opened 2022-03-25
Added pytest_args option to `additional_repos`

bmbouter:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ The following settings are stored in `template_config.yml`.
                         org (optional): github organization
                         revision (optional): the git commit hash to install.
                                              "Required PR" in a commit message will override this.
+                        pytest_args (optional): string to by passed to a pytest call.
                         branch: the git branch of the plugin. applies to non-tagged GHA jobs, such as PRs & cron jobs.
                         pip_version_specifier: A pip version specifier for when the gets installed from PyPI.
                                                Applies to TAGGED (release) jobs.
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -196,16 +196,34 @@ else
 {% if parallel_test_workers == 0 %}
     if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
+
+    {%- for item in additional_repos if item.pytest_args | default(false) %}
+        pytest -v -r sx --color=yes --pyargs {{ item.pytest_args }}
+    {%- endfor %}
     else
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not nightly"
+
+    {%- for item in additional_repos if item.pytest_args | default(false) %}
+        pytest -v -r sx --color=yes --pyargs {{ item.pytest_args }} -m "not nightly"
+    {%- endfor %}
     fi
 {% else %}
     if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs {{ plugin_snake }}.tests.functional -m parallel -n {{ parallel_test_workers }}
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional -m "not parallel"
+
+    {%- for item in additional_repos if item.pytest_args | default(false) %}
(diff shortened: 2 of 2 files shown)


bmbouter:
I don't love this approach because it's very specific to the pulpcore use case. Maybe we should be auto-running the tests from the `additional_repos` data, or make a new section in the plugin_template?

fao89 (approved):

dkliban:
This approach is fine. It's the only place where we run additional plugin's tests.

fao89:
> I don't love this approach because it's very specific to the pulpcore use case. Maybe we should be auto-running the tests from the `additional_repos` data, or make a new section in the plugin_template?

you can have it on `func_test_script.sh`
https://github.com/pulp/pulp_file/blob/main/.github/workflows/scripts/func_test_script.sh

dkliban (approved):

bmbouter:
Using `additional_repos` gets complicated when you have pulp_ansible using it like this https://github.com/pulp/pulp_ansible/blob/main/template_config.yml#L6-L10 So if we want to make this more generalized I would make a new section in template_config like `additional_functional_tests`?

bmbouter:
I could have it be in `func_test_script.sh` but I'm wondering, does pulp_file need to run the certguard tests, and does pulpcore also need to run the pulp-certguard tests? If so it would be better to make this feature more general. I'm not sure. Maybe we should just use this for now and do the bigger generalization later.

bmbouter:
Let's not merge this, I'm working on the more generalized version.

bmbouter:
Check this out https://github.com/pulp/pulpcore/pull/2432

README.md
```diff
@@ -54,6 +54,7 @@ The following settings are stored in `template_config.yml`.
                         org (optional): github organization
                         revision (optional): the git commit hash to install.
                                              "Required PR" in a commit message will override this.
+                        pytest_pyargs (optional): string to by passed to pytest via pyargs
```
mdellweg, line 57: I'd expect a list of plugins from which we also run the functional tests.

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -196,16 +196,34 @@ else
 {% if parallel_test_workers == 0 %}
     if [[ "$GITHUB_WORKFLOW" == "{{ plugin_camel_short | default("Pulp") }} Nightly CI/CD" ]]; then
         pytest -v -r sx --color=yes --pyargs {{ plugin_snake }}.tests.functional
+
+    {%- for item in additional_repos if item.pytest_pyargs | default(false) %}
+        pytest -v -r sx --color=yes --pyargs {{ item.pytest_pyargs }}
+    {%- endfor %}
```
mdellweg, line 202: Can we combine these in a single pytest invocation? Maybe add an option to stop trying after 20 failed tests.

README.md
mdellweg: Ah, I misread. The context in the diff was not enough without expanding.

mdellweg: But also to me that sounds like a feature switch:
`--pyargs              try to interpret all arguments as python packages.`

So technically we only add arguments to pytest, and the switch tells pytest to interpret _everything_ as python modules.

bmbouter: This is better I think because it doesn't rely on the assumption that all tests are in a canonical place. Also it's simpler because I don't have to build that string up in code, the data can just have it.

templates/github/.github/workflows/scripts/script.sh.j2
bmbouter: I think having them separate is kind of nice, that way if there is a failure you have big clear blocks showing you exactly where it is instead of one big run. The stopping after some time sounds nice but its not really the goal of this work.

README.md
mdellweg: I agree (see my second comment) i just don't really dig the name. Is it `test_module`?

templates/github/.github/workflows/scripts/script.sh.j2
mdellweg: The thing is that we stop processing after the first failed block. If we can combine test suites in one pytest instantiation, you would see all the failed tests you need to fix to make the CI green again.
Also we loose half of the "--parallel" benefit.
And now i realize the parallel tests suffer from the same restriction. They cannot be called in the same run with the non parallel ones.
I'm on the fence here.

README.md
mdellweg: What i like about this approach is that you could specify a subset of tests to run with pulpcore. But i'd rather have a `@pytest.mark.run_with_core` for this to add visibility.

bmbouter: I named it with `pyargs` because it's possible we may want to add a section that doesn't use `pyargs` perhaps? Like passing the string directly to pytest.

bmbouter:
I renamed this to `pytest_args` and updated the README.

bmbouter:
@mdellweg I'm going to merge this, but lmk if there is more that I should do to it. I suspect I'll be improving this machinery more over time.

PR #604, opened 2022-03-28
Pin click to run black

fao89:
[noissue]

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -27,7 +27,8 @@ jobs:
       - name: Lint plugin_template
         working-directory: plugin_template
         run: |
-          pip3 install black flake8
+          # Unpin click when https://github.com/psf/black/issues/2964 is solved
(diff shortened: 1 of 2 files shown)


bmbouter:
I think this is right. Do we have an issue open with the click community tracking their resolution of it?

fao89:
> I think this is right. Do we have an issue open with the click community tracking their resolution of it?

we have it on black:
https://github.com/psf/black/issues/2964

.github/workflows/bootstrap.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Lint plugin_template
         working-directory: plugin_template
         run: |
-          pip3 install black flake8
+          pip3 install black flake8 "click<8.1"
```
bmbouter, line 30: Wonderful. Can you put that link in the comment of this line and also the line below it. Something like `Unpin click when <link here> is resolved.

bmbouter (approved):
Thank you!

PR #606, opened 2022-03-31
Clean python-client-docs.tar on release job

fao89:
[noissue]


mdellweg (approved):
How much cleaning should be done here? would a `git clean ...` of some sort work?

fao89:
> How much cleaning should be done here? would a `git clean ...` of some sort work?

it was the last piece for cleaning

PR #612, opened 2022-04-12
Switches the unittest runner to use pytest

dkliban:
[noissue]


bmbouter (approved):
This is great, thank you!

PR #613, opened 2022-04-12
Check towncrier entries on docs job

### LINKED ISSUE #610: docs CI job should check that the changelog can be published using sphinx
I just ran into a problem with the changelog during release workflow. In my case there were two changelog entries that were  causing a problem. CI for every PR should try to run towncrier to generate a changelog and then build the docs using sphinx.

fao89:
closes #610

### DIFF
--- a/CHANGES/610.feature
+++ b/CHANGES/610.feature
@@ -1,0 +1,1 @@
+Check towncrier entries on docs job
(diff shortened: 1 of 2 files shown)


fao89:
Tested at:
https://github.com/pulp/pulp_ansible/pull/910

https://github.com/pulp/pulp_ansible/runs/5989297052?check_suite_focus=true

dkliban (approved):

PR #620, opened 2022-04-20
Fix commit validation for cherry-picked commits

### LINKED ISSUE #619: Cherry-pick automation does not work when merge commits are used
It looks like when using the patchback cherry-pick automation, the merge commit will be cherry picked.

Since the merge commit does not contain the issue reference in the commit message, the so opened PR will then fail the commit message validation as follows: https://github.com/pulp/pulp_deb/pull/483

* One possible solution is to change the pulp_deb repo settings to stop using merge commits (and use manual cherry-picks until then).
* Another possibility would be to always add the issue reference in the merge commit message going forwards.
* The final option would be to have the automation cherry-pick not the merge commit, but the individual commits on the PR being backported.

fao89:
closes #619

### DIFF
--- a/CHANGES/619.bugfix
+++ b/CHANGES/619.bugfix
@@ -1,0 +1,1 @@
+Fix commit validation for cherry-picked commits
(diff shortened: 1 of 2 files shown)


fao89:
@quba42 wdyt?

quba42:
Tested the fix here: https://github.com/pulp/pulp_deb/pull/483 (linting is now green)

PR #629, opened 2022-04-26
Fix the upgrade tests, once more

bmbouter:
[noissue]


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -155,9 +155,8 @@ if [[ "$TEST" == "upgrade" ]]; then
   # Let's reinstall pulpcore so we can ensure we have the correct dependencies
   cd ../pulpcore
   git checkout -f ci_upgrade_test
-  pip install --upgrade --force-reinstall .
   cd ..
-  pip install --upgrade ../pulp-cli ../pulp-smash
+  pip install --upgrade --force-reinstall pulpcore ../pulp-cli ../pulp-smash
```
fao89, line 159: I think it should be
```suggestion
  pip install --upgrade --force-reinstall ../pulpcore ../pulp-cli ../pulp-smash
```

bmbouter: I'm trying your suggestion now here and also https://github.com/pulp/pulp_file/pull/694

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -155,9 +155,8 @@ if [[ "$TEST" == "upgrade" ]]; then
   # Let's reinstall pulpcore so we can ensure we have the correct dependencies
   cd ../pulpcore
```
fao89, line 156: I think we need to remove this line

dkliban (approved):

PR #632, opened 2022-05-02
Add bandit

fao89:
https://bandit.readthedocs.io/en/latest/index.html

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -40,7 +40,12 @@ jobs:
 
         # dev_requirements contains tools needed for flake8, etc.
       - name: Install requirements
-        run: pip3 install -r dev_requirements.txt
+        run: |
+          pip3 install -r dev_requirements.txt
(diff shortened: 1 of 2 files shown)


ggainey:
I think I'd like to see us run bandit against projects, clean up what really needs to be cleaned up, determine which things are "ok" (e.g., using non-crypto-secure random-number-generators is not a problem for things like "choose one random entity in a test"), and set up a global .bandit file. If this merged right now, the next global template-update would break stop us dead in our tracks, and a lot of the failures in the affected check here are not "stop the proesses" problems.

bmbouter:
What about posting on discourse about this and asking each of the plugins to report their readiness there? Having a simple example of what to run locally would probably be helpful. Then a switchover date could be coordinated to merge this as a new CI requirement.

mdellweg:
Actually, looking at the bandit complaints here, i am kind of disappointed. I know, this is only the plugin template, but we are probably going to classify _all_ of them valid use.

PR #635, opened 2022-05-05
New option to also run marked tests from pulpcore

bmbouter:
[noissue]


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -32,13 +32,6 @@ if [[ "$TEST" = "docs" ]]; then
   make PULP_URL="$PULP_URL" diagrams html
   tar -cvf docs.tar ./_build
   cd ..
-{%- if check_openapi_schema %}
```
fao89, line 35: you can drop it from here: https://github.com/pulp/plugin_template/blob/main/plugin-template#L28

bmbouter: good idea!

plugin-template
```diff
@@ -62,6 +61,7 @@ DEFAULT_SETTINGS = {
     "python_version": "3.8",
     "release_user": "pulpbot",
     "release_email": "pulp-infra@redhat.com",
+    "run_pulpcore_tests_for_plugins": True,
```
fao89, line 64: 👍🏽

fao89 (approved):

PR #637, opened 2022-05-06
Stops pinning sphinx in doc_requirements.txt

dkliban:
[noissue]


bmbouter (approved):
It looks like you fixed it!

PR #643, opened 2022-05-13
Install additional plugins to provide pytest parts

mdellweg:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -21,7 +21,7 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   pip install -r doc_requirements.txt
 fi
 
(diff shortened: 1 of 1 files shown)


fao89 (approved):

bmbouter:
I'm glad this is here for now. I'm believing the longer term plan is to have the requirements files handle all of these installs and not have their need expressed via template_config vars. @mdellweg does that make sense to you. Also @dkliban FYI re your refactoring I think will need to handle this change specifically.

mdellweg:
> I'm glad this is here for now. I'm believing the longer term plan is to have the requirements files handle all of these installs and not have their need expressed via template_config vars. @mdellweg does that make sense to you. Also @dkliban FYI re your refactoring I think will need to handle this change specifically.

If you are thinking about test_requirements here, i'm with you. But we might still need to install the _other_ plugins by hand from source.

PR #653, opened 2022-06-06
Revert "Install additional plugins to provide pytest parts"

fao89:
Reverts pulp/plugin_template#643

it breaks when pulp-rpm is listed on additional plugins, as pulp-rpm can't be installed on ubuntu
https://github.com/pulp/pulp-2to3-migration/runs/6749755516?check_suite_focus=true
https://github.com/pulp/pulp-2to3-migration/blob/main/template_config.yml#L12

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -21,7 +21,7 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   pip install -r doc_requirements.txt
 fi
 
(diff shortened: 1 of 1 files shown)


dralley (approved):

dralley:
Where are the plugins being installed, inside the pulp environment or outside in the test environment?  We definitely do want these installed to test the migration plugin since the migration plugin depends on their presence

ipanova (approved):

bmbouter:
We need this for pulpcore and pulp_file, I believe we're waiting on the work from @dkliban to resolve this perhaps? Please let's not revert this.

dkliban:
@fao89 Let's not merge this. I believe we just want to add `--no-deps` to the `pip install` command.

Issue #642, opened 2022-05-12
run functional tests inside the container running pulp


dkliban:
All functional tests should be run from inside the container. The PYTHONPATH should no longer need to be manipulated. Functional test requirements should only be installed inside the container.

PR #657, opened 2022-06-08
Runs functional tests inside the container running Pulp

dkliban:
closes: #642

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -56,34 +56,35 @@ if [[ "$TEST" == "plugin-from-pypi" ]]; then
   git checkout ${COMPONENT_VERSION} -- {{ plugin_name }}/tests/
 fi
 
+echo "machine pulp
+login admin
+password password
+" | cmd_stdin_prefix bash -c "cat > /root/.netrc"
+
+cat unittest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/unittest_requirements.txt"
+cat functest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/functest_requirements.txt"
+cmd_prefix pip3 install -r /tmp/unittest_requirements.txt
+cmd_prefix pip3 install -r /tmp/functest_requirements.txt
+cmd_prefix pip3 install --upgrade ../pulp-smash
+
 cd ../pulp-openapi-generator
 {%- if plugin_name != 'pulpcore' %}
+./generate.sh {{ plugin_snake }} python
+cmd_prefix pip3 install /root/pulp-openapi-generator/{{ plugin_snake }}-client
+sudo rm -rf ./{{ plugin_snake }}-client
 ./generate.sh pulpcore python
-pip install ./pulpcore-client
-rm -rf ./pulpcore-client
-if [[ "$TEST" = 'bindings' ]]; then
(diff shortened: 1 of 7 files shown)


templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -7,6 +7,12 @@ ADD {{ item.source }} {{ item.source }}
 {% endif %}
 {% endfor %}
 
+{%- for item in plugins -%}
+{%- if item.name == "pulp_container" -%}
+RUN dnf install -y podman
```
fao89, line 12: why not having it by default?

templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -7,6 +7,12 @@ ADD {{ item.source }} {{ item.source }}
 {% endif %}
 {% endfor %}
 
+{%- for item in plugins -%}
```
newswangerd, line 10: @dkliban it seemed like (and I might be wrong about this) that the pip dependencies are installed each time CI is run. Would it be possible to install the basic pulpcore requirements [here](https://github.com/pulp/pulp-oci-images/blob/latest/pulp_ci_centos/Containerfile), and then run `pip install --upgrade` when the plugins are installed in CI?

A large chuck of the CI time is just downloading the pip packages during the container build. Would be nice to make that run faster.

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -56,27 +56,36 @@ if [[ "$TEST" == "plugin-from-pypi" ]]; then
   git checkout ${COMPONENT_VERSION} -- {{ plugin_name }}/tests/
 fi
 
+cat unittest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/unittest_requirements.txt"
+cat functest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/functest_requirements.txt"
+cmd_prefix pip3 install -r /tmp/unittest_requirements.txt
```
newswangerd, line 61: If I'm not mistaken, cmd_prefix just runs `docker exec`. Would it make sense to move these steps where you're just installing extra requirements in the container to the `Containerfile`?

templates/github/.ci/ansible/Containerfile.j2.copy
dkliban: That is a good idea.  I would like to implement it as a separate change from this.

templates/github/.ci/ansible/Containerfile.j2.copy
dkliban: Good idea. I opened a PR for this. https://github.com/pulp/pulp-oci-images/pull/203

templates/github/.github/workflows/scripts/script.sh.j2
dkliban: Yes, that would make sense. I'll make that change.

dkliban: I tried implementing this and it's not going to work without doing a lot of refactoring to how we build the container. The current Containerfile template doesn't care if the plugin is being installed from source (PR time) or from a prebuilt wheel (for release pipeline). When the wheel is being used, the functest_requirements.txt and unittest_requirements.txt are not available during the built.

fao89 (approved):
LGTM

PR #664, opened 2022-06-29
Installs cli test requirements if test_cli is True.

dkliban:
[noissue]


gerrod3 (approved):

gerrod3 (approved):
Even more simple, nice.

PR #666, opened 2022-07-05
Lint requirements

fao89:
[noissue]

### DIFF
--- a/templates/github/.ci/scripts/upper_bound.py.j2
+++ b/templates/github/.ci/scripts/upper_bound.py.j2
@@ -1,3 +1,4 @@
+import warnings
 from pkg_resources import Requirement
 
 packages = []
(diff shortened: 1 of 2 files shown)


bmbouter:
@dralley @mdellweg @ipanova @lubosmj @quba42 @gerrod3 @dkliban @ggainey  I'm pinging all the leads from plugins as this change when merged and applied will likely break your CI. Really though it's moving a release-time check to a PR-time check so you'd have to do this anyway.

When specifying upper bounds I recommend having the broadest compatibility possible, so instead of using `~=` use as broad a range as you support. Then dependabot will continue to bump your upper bound as needed, so this change should live nicely on `main` branches. Any questions or comments are welcome.

quba42:
As far as I can tell this just means: "Add `,<3.25` to the pulpcore requirement on `main` branch".

Do I need to understand any deeper reasons as to why we are doing this now?

bmbouter:
> As far as I can tell this just means: "Add `,<3.25` to the pulpcore requirement on `main` branch for plugins".
It's not just pulpcore, it's really all the dependencies named in the dependency file.

> 
> Do I need to understand any deeper reasons as to why we are doing this now?

In terms of "why now?" I'd say there are two motivators. 1) We couldn't before (at least for pulpcore) without having [a clearly established breaking change policy](https://discourse.pulpproject.org/t/what-to-do-to-reduce-unecessary-plugin-compatibility-releases/236/18) in the plugin API. 2) We are experiencing a practical problem where the "check that all deps have a declared upper bound" is not enforced at PR-merge-time but is at release-time. This causes many releases to fail, creating extra work.

PR #667, opened 2022-07-10
[bugfix] Make lint_requirements optional

rochacbruno:
Reason: By default this check `upper_bounds.py` looks for a `requirements.txt`
and galaxy_ng doesn´t have that file and it was failing teh CI.

[noissue]


rochacbruno:
Other option is to change the `upper_bounds` script and check for the existance of `requirements.txt` and if not exists bypass the check.

fao89 (approved):

PR #674, opened 2022-08-03
Tweet releases

fao89:
[noissue]

### DIFF
--- a/templates/github/.ci/scripts/tweet.py.j2
+++ b/templates/github/.ci/scripts/tweet.py.j2
@@ -1,0 +1,16 @@
+import os
+import sys
+from tweepy import Client
+
+release_version = sys.argv[1]
+if release_version.endswith(".0"):
+    client = Client(
+        consumer_key=os.getenv("TWITTER_API_KEY"),
+        consumer_secret=os.getenv("TWITTER_API_KEY_SECRET"),
+        access_token=os.getenv("TWITTER_ACCESS_TOKEN"),
+        access_token_secret=os.getenv("TWITTER_ACCESS_TOKEN_SECRET"),
+    )
+    link = "https://docs.pulpproject.org/{{ plugin_snake }}/changes.html"
+    msg = f"{{ plugin_name }}-{release_version} - Check out for more details: {link}"
+    release_msg = f"Hey! We've just released {msg}"
+    client.create_tweet(text=release_msg)
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -208,7 +208,7 @@ jobs:
 
       {{ set_secrets() | indent(6) }}
 
-      {{ install_python_deps("gitpython python-redmine requests packaging") | indent(6) }}
+      {{ install_python_deps("gitpython python-redmine requests packaging tweepy") | indent(6) }}
 
       - name: Push branch and tag to GitHub
         run: bash .github/workflows/scripts/push_branch_and_tag_to_github.sh {{ "${{ github.event.inputs.release }}" }}
@@ -279,6 +279,10 @@ jobs:
           echo "GH Issues $GH_ISSUES"
           python .ci/scripts/update_github.py
       {%- endif %}
+
+      - name: Tweet
+        continue-on-error: true
+        run: python .ci/scripts/tweet.py {{ "${{ github.event.inputs.release }}" }}
 
       - name: Create release on GitHub
         run: bash .github/workflows/scripts/create_release_from_tag.sh {{ "${{ github.event.inputs.release }}" }}


templates/github/.ci/scripts/tweet.py.j2
```diff
@@ -0,0 +1,15 @@
+import os
+import sys
+from tweepy import Client
+
+release_version = sys.argv[1]
+client = Client(
+    consumer_key=os.getenv("TWITTER_API_KEY"),
+    consumer_secret=os.getenv("TWITTER_API_KEY_SECRET"),
+    access_token=os.getenv("TWITTER_ACCESS_TOKEN"),
+    access_token_secret=os.getenv("TWITTER_ACCESS_TOKEN_SECRET"),
```
fao89, line 10: we need to set these keys on github

RCMariko: @bmbouter has the info we need. I can set up the keys and he may also be able to do that - but we'll have to wait til he is back.

goosemania:
Thank for putting this together, I need to think more about it.

I think we likely want to post only for Y releases and when it makes sense. E.g. if it's a pulpcore release with some breaking changes for plugins, we probably do not want to tweet about it immediately but wait for plugins. And when plugins are released, I guess it's pretty annoying to get N tweets fro each plugin, rather then one consolidated one for pulpcore and compatible plugins :thinking: 

I like the idea of posting a tweet easily though. Maybe a manual job which asks for a message to provide :thinking: will do the trick.

bmbouter:
+1 to tweeting for Y-streams only. I haven't looked at the code to see what it actually does. I don't know how to make it fancier to wait for pulpcore+plugin releases, so I'd settle for just any Y release. I think we should auto-tweet something because with fewer people we need more automation.

templates/github/.ci/scripts/tweet.py.j2
```diff
@@ -0,0 +1,16 @@
+import os
+import sys
+from tweepy import Client
+
+release_version = sys.argv[1]
+if release_version.endswith(".0"):
```
fao89, line 6: I already modified the code to be Y-stream only

ipanova:
> Thanks for putting this together, I need to think more about it.
> 
> I think we likely want to post only for Y releases and when it makes sense. E.g. if it's a pulpcore release with some breaking changes for plugins, we probably do not want to tweet about it immediately but wait for plugins. And when plugins are released, I guess it's pretty annoying to get N tweets fro each plugin, rather then one consolidated one for pulpcore and compatible plugins thinking
We have every 5th release a breaking one so that should not be that big of the problem, I think.
> 
> I like the idea of posting a tweet easily though. Maybe a manual job which asks for a message to provide thinking will do the trick.

This is almost like going and doing the tweet yourself :shrug:

ipanova:
> +1 to tweeting for Y-streams only. I haven't looked at the code to see what it actually does. I don't know how to make it fancier to wait for pulpcore+plugin releases, so I'd settle for just any Y release. I think we should auto-tweet something because with fewer people we need more automation.

+1

goosemania:
> 
> > I like the idea of posting a tweet easily though. Maybe a manual job which asks for a message to provide thinking will do the trick.
> 
> This is almost like going and doing the tweet yourself shrug

Yup, and it's fine since folks do not need credentials or go anywhere else apart from github. IMO, reduces friction.

templates/github/.ci/scripts/tweet.py.j2
```diff
@@ -0,0 +1,16 @@
...
+    client = Client(
+        consumer_key=os.getenv("TWITTER_API_KEY"),
+        consumer_secret=os.getenv("TWITTER_API_KEY_SECRET"),
+        access_token=os.getenv("TWITTER_ACCESS_TOKEN"),
+        access_token_secret=os.getenv("TWITTER_ACCESS_TOKEN_SECRET"),
+    )
+    link = "https://docs.pulpproject.org/{{ plugin_snake }}/changes.html"
+    msg = f"{{ plugin_name }}-{release_version} - Check it out at: {link}"
```
ipanova, line 14: "Check out for more details :...."

templates/github/.ci/scripts/tweet.py.j2
```diff
@@ -0,0 +1,16 @@
...
+
+release_version = sys.argv[1]
+if release_version.endswith(".0"):
+    client = Client(
+        consumer_key=os.getenv("TWITTER_API_KEY"),
+        consumer_secret=os.getenv("TWITTER_API_KEY_SECRET"),
+        access_token=os.getenv("TWITTER_ACCESS_TOKEN"),
+        access_token_secret=os.getenv("TWITTER_ACCESS_TOKEN_SECRET"),
+    )
+    link = "https://docs.pulpproject.org/{{ plugin_snake }}/changes.html"
```
ipanova, line 13: did you mean plugin_name?

fao89: nope, plugin_snake for ensuring the _
https://docs.pulpproject.org/pulp_2to3_migration/

ipanova (commented):
looks good to me!

ipanova (approved):

templates/github/.ci/scripts/tweet.py.j2
```diff
@@ -0,0 +1,16 @@
+import os
+import sys
+from tweepy import Client
```
bmbouter, line 3: How is this dependency being installed?

bmbouter: Oh I see it further down. Great!

fao89:
> @fao89 thank you for this. Please merge when you feel all comments are addressed (if not already). Please let me know if I need to configure any credentials anywhere for this to work.

we need to set these secrets on pulp org:
- TWITTER_API_KEY
- TWITTER_API_KEY_SECRET
- TWITTER_ACCESS_TOKEN
- TWITTER_ACCESS_TOKEN_SECRET

PR #677, opened 2022-08-11
Install pulp-cli not in editable mode

dkliban:
For some reason on GHA pulp-cli installed in editable mode is causing python to not be
able to import `pulpcore.cli`.

[noissue]


bmbouter (approved):
awesome!

dralley (approved):

mdellweg:
I would really like to understand "some reason". But being in editable mode was unneeded anyway.

PR #680, opened 2022-08-23
Adjustments on release job

fao89:
This is what I need to do to release pulp_ansible 0.13.4

### DIFF
--- a/templates/github/.ci/ansible/Containerfile.j2.copy
+++ b/templates/github/.ci/ansible/Containerfile.j2.copy
@@ -10,7 +10,7 @@ ADD {{ item.source }} {{ item.source }}
 # Install python packages
 # Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
-RUN pip3 install \
+RUN pip3 install --no-binary pulpcore \
 {%- if stream_test | default(false) -%}
 {{ " " }}django-storages[sftp]
 {%- endif -%}
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -46,8 +46,6 @@ PULPCORE=./pulpcore
 {%- endif %}
 if [[ "$TEST" == "plugin-from-pypi" ]]; then
   PLUGIN_NAME={{ plugin_name }}
-elif [[ "${RELEASE_WORKFLOW:-false}" == "true" ]]; then
-  PLUGIN_NAME=./{{ plugin_name }}/dist/{{ plugin_snake }}-$PLUGIN_VERSION-py3-none-any.whl
 else
   PLUGIN_NAME=./{{ plugin_name }}
 fi


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -37,8 +37,6 @@ PULPCORE=./pulpcore
 {%- endif %}
 if [[ "$TEST" == "plugin-from-pypi" ]]; then
   PLUGIN_NAME={{ plugin_name }}
-elif [[ "${RELEASE_WORKFLOW:-false}" == "true" ]]; then
-  PLUGIN_NAME=./{{ plugin_name }}/dist/{{ plugin_snake }}-$PLUGIN_VERSION-py3-none-any.whl
```
fao89, line 41: whl does not consider MANIFEST.in
https://stackoverflow.com/questions/24347450/how-do-you-add-additional-files-to-a-wheel

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -199,7 +197,11 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
 {%- if plugin_name == 'pulpcore' %}
   pip install psycopg2-binary
 {%- else %}
-  cmd_prefix bash -c "cd pulpcore; pip install -r doc_requirements.txt"
+  if [[ "${RELEASE_WORKFLOW:-false}" == "true" ]]; then
+    cmd_prefix bash -c "pip install -r https://raw.githubusercontent.com/pulp/pulpcore/{{ pulpcore_branch }}/doc_requirements.txt"
```
fao89, line 201: ugly hack

templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -10,7 +10,7 @@ ADD {{ item.source }} {{ item.source }}
 # Install python packages
 # Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
-RUN pip3 install \
+RUN pip3 install --no-binary=:all: \
```
fao89, line 13: prefer tar.gz

fao89: actually it breaks, because it can't install cryptography:
https://github.com/pulp/plugin_template/runs/7976724965?check_suite_focus=true#step:8:905

templates/github/.github/workflows/scripts/install.sh.j2
bmbouter: I think we generally don't want additional files to be included in wheels at all. It's designed to just be application code (my understanding). I use the term application code to indicate disinclusion of test python code (which is still python code). The one thing every distribution type (including wheel) needs to include is the license file.

templates/github/.github/workflows/scripts/install.sh.j2
bmbouter: I think we want to have pip install the plugin under test using the `--no-binary` option. As an example for pulpcore I think that would be `pip install --no-binary pulpcore pulpcore`

bmbouter (changes requested):
@fao89 wdyt about these changes?

templates/github/.github/workflows/scripts/install.sh.j2
fao89: when we try to install pulpocre this way, the CI fails, as it can't install cryptography
https://github.com/pulp/plugin_template/pull/680#discussion_r952783847

templates/github/.ci/ansible/Containerfile.j2.copy
fao89: let me try it

bmbouter:
@fao89 it has conflicts now

bmbouter:
@fao89 wdyt of this PR now?

fao89:
> @fao89 wdyt of this PR now?

the main problem that motivated this PR is solved, as the docs tests are running outside of the container,
but plugins may face issues with whl x MANIFEST.in in the future

As the releases are unblocked now, I think we can wait for @dkliban review

PR #681, opened 2022-08-24
Revert docs test inside container

fao89:


bmbouter:
@fao89 what should we do about the failing first check?

fao89:
> @fao89 what should we do about the failing first check?

ignore

PR #686, opened 2022-08-31
Add flake8_ignore to ignore additional files

mdellweg:
[noissue]


fao89 (approved):

bmbouter (approved):
I am planning to have protobuf stuff become its own python library shipped through pypi so it won't have to be checked in to any of our other repos, but between now and then this is a great solution!

PR #691, opened 2022-09-13
Remove allow_async_unsafe

dralley:
This requires that we stop updating the CI on the 3.14 branch, which I think is actually perfectly fine at this point.


mdellweg:
I would rather stop using it before removing it. But maybe yes it _is_ the right moment.

dralley:
We can't stop using it on 3.14 unfortunately

mdellweg (approved):

PR #692, opened 2022-09-14
Add extra settings for s6 scenario

mdellweg:
[noissue]


templates/github/.ci/ansible/settings.py.j2.copy
```diff
@@ -20,6 +20,12 @@ API_ROOT = {{ api_root | repr }}
 {% endfor %}
 {% endif %}
 
+{% if pulp_scenario_settings is defined and pulp_scenario_settings %}
+{% for key, value in pulp_scenario_settings.items() %}
+{{ key | upper }} = {{ value | repr }}
+{% endfor %}
+{% endif %}
+
```
fao89, line 28: why not just use `pulp_settings`?

mdellweg: Because i want different values and different settings in different scenarios. That way we can test the feature being on and off.

fao89: only now I realize it is only applied to the S3 job,
I think the name is misleading, s6 sounds like a specific single container config

mdellweg: I'll add the same logic for the other scenarios then.

fao89 (approved):
nice!

plugin-template
```diff
@@ -52,6 +52,9 @@ DEFAULT_SETTINGS = {
     "plugin_snake": None,
     "publish_docs_to_pulpprojectdotorg": False,
     "pulp_settings": None,
+    "pulp_settings_azure": None,
+    "pulp_settings_s6": None,
```
bmbouter, line 56: You meant this not `s3` right?

mdellweg: :facepalm: Damn it!

PR #693, opened 2022-09-15
Use specified branches for ci updates

### LINKED ISSUE #684: Workflow for updating CI configs needs to only update a set of branches
Right now the workflow tries to update all branches going back to a certain one. The workflow should only update branches that are part of a list of supported branches.

mdellweg:
fixes #684

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,8 +138,7 @@ The following settings are stored in `template_config.yml`.
                         'github'. To switch from Redmine to GitHub use the --migrate-github-issues
                         option.
 
-  keep_ci_update_since_branch
-                        Keep CI updated for all starting from the specified branch.
+  ci_update_branches    Specify the branches that should receive regular CI updates.
 
   docs_test             Include a CI build for testing the 'make html' command for sphinx docs.
 
--- a/plugin-template
+++ b/plugin-template
@@ -37,7 +37,7 @@ DEFAULT_SETTINGS = {
     "docs_test": True,
     "github_org": "pulp",
     "issue_tracker": "github",
-    "keep_ci_update_since_branch": None,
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -138,8 +138,9 @@ The following settings are stored in `template_config.yml`.
                         'github'. To switch from Redmine to GitHub use the --migrate-github-issues
                         option.
 
+  ci_update_branches    Specify the branches that should receive regular CI updates.
   keep_ci_update_since_branch
-                        Keep CI updated for all starting from the specified branch.
+                        Keep CI updated for all starting from the specified branch. DEPRECATED
```
bmbouter, line 143: Can we just remove it? The worst case is plugin writers will not have some branches receive updates automatically, and then they would be like "oh I need to switch".

templates/github/.ci/scripts/update_ci_branches.py.j2
```diff
@@ -1,3 +1,5 @@
+{% include 'header.j2' %}
```
bmbouter, line 1: What does this do?

mdellweg: I tells you: "Can't touch this!"

README.md
mdellweg: I'm trying to follow the principle of least surprise.

plugin-template
```diff
@@ -257,6 +257,10 @@ def main():
                         if key not in config:
                             config[key] = value
                             write_new_config = True
+                    # remove deprecated options
+                    if "keep_ci_update_since_branch" in config:
```
bmbouter, line 261: Awww yeah!

PR #697, opened 2022-09-21
Use 4 pulp worker processes in the container

mikedep333:
rather than the default of 2.

This is because the github actions CI only has 2 cores, even though
parallel_test_workers defaults to 8 and is configurable.

[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -119,6 +119,8 @@ services:
       - ./ssh:/keys/
       - ~/.config:/root/.config
       - ../../../pulp-openapi-generator:/root/pulp-openapi-generator
+    env:
+      PULP_WORKERS: "4"
 VARSYAML
 
 cat >> vars/main.yaml << VARSYAML


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -119,6 +119,8 @@ services:
       - ./ssh:/keys/
       - ~/.config:/root/.config
       - ../../../pulp-openapi-generator:/root/pulp-openapi-generator
+    env:
+      PULP_WORKERS: "9"
```
fao89, line 123: why 9?

fao89 (approved):

templates/github/.github/workflows/scripts/install.sh.j2
mikedep333: I saw 9 listed here: https://github.com/pulp/pulp-oci-images/issues/225

mikedep333: Actually, I updated it to be variable. Based on the templated value of parallel_test_workers.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -110,6 +110,8 @@ plugins:
 VARSYAML
 {%- endif %}
 
+PULP_WORKERS=$(({{ parallel_test_workers }}+1))
```
fao89, line 113: sweet!

fao89 (approved):

Issue #698, opened 2022-09-29
Multiple functional tests fail with the limited pulp user account in the container


mikedep333:
Now that the pulp services run as a limited user in the container, rather than as root, our tests that run as root fail.

Presumably because they create files on disk as root.

PR #699, opened 2022-09-29
Fix tests failing with the limited user pulp account

mikedep333:
in the container

fixes: #698

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -9,6 +9,7 @@ REPO_ROOT="$PWD"
 
 set -mveuo pipefail
 
+shopt -s expand_aliases
 source .github/workflows/scripts/utils.sh
 
 export POST_SCRIPT=$PWD/.github/workflows/scripts/post_script.sh
@@ -59,8 +60,8 @@ fi
 echo "machine pulp
 login admin
 password password
-" | cmd_stdin_prefix bash -c "cat > /root/.netrc"
-cmd_stdin_prefix bash -c "chmod og-rw /root/.netrc"
+" | cmd_stdin_prefix bash -c "cat > ~pulp/.netrc"
+cmd_stdin_prefix bash -c "chmod og-rw ~pulp/.netrc"
 
 cat unittest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/unittest_requirements.txt"
 cat functest_requirements.txt | cmd_stdin_prefix bash -c "cat > /tmp/functest_requirements.txt"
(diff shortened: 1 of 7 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -59,8 +59,8 @@ fi
 echo "machine pulp
 login admin
 password password
-" | cmd_stdin_prefix bash -c "cat > /root/.netrc"
-cmd_stdin_prefix bash -c "chmod og-rw /root/.netrc"
+" | cmd_stdin_prefix bash -c "cat > ~pulp/.netrc"
+cmd_stdin_prefix bash -c "chmod og-rw ~pulp/.netrc"
```
mdellweg, line 63: This is because we run the tests as the same user pulp?

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -89,17 +89,17 @@ cmd_prefix bash -c "cat /etc/pulp/certs/pulp_webserver.crt  | tee -a "$CERTIFI"
...
 
 # Run unit tests.
-cmd_prefix bash -c "PULP_DATABASES__default__USER=postgres pytest -v -r sx --color=yes -p no:pulpcore --pyargs {{ plugin_snake }}.tests.unit"
+cmd_user_prefix bash -c "PULP_DATABASES__default__USER=postgres pytest -v -r sx --color=yes -p no:pulpcore --pyargs {{ plugin_snake }}.tests.unit"
 
 # Run functional tests
 if [[ "$TEST" == "performance" ]]; then
   if [[ -z ${PERFORMANCE_TEST+x} ]]; then
-    cmd_prefix bash -c "pytest -vv -r sx --color=yes --pyargs --capture=no --durations=0 {{ plugin_snake }}.tests.performance"
+    cmd_user_prefix bash -c "pytest -vv -r sx --color=yes --pyargs --capture=no --durations=0 {{ plugin_snake }}.tests.performance"
```
mdellweg, line 100: I guess our tests are not ready to run as a different user...

mdellweg (approved):

.github/workflows/bootstrap.yaml
```diff
@@ -212,6 +212,10 @@ jobs:
           docker exec pulp ls -latr /etc/yum.repos.d/ || true
           docker exec pulp cat /etc/yum.repos.d/* || true
           docker exec pulp pip3 list
+
+      - name: tmate
+        if: failure()
+        uses: mxschmitt/action-tmate@v3
```
fao89, line 218:
```suggestion

```

templates/github/.github/workflows/scripts/script.sh.j2
mikedep333: Yes

templates/github/.github/workflows/scripts/script.sh.j2
mikedep333: We are now :)

mikedep333: @dkliban and I have been discussing this. We think it's best to run the tests as the pulp user, including on the oci dev env.

We were able to fix some issues (e.g., precreating /var/lib/pulp/media/artifact with pulp as the owner, rather than root), but still thought that this approach is better.

PR #700, opened 2022-10-03
Revert cmd_prefix and similar from aliases to functions

mikedep333:
Allevitates needs for scripts like post_script.sh to allow aliases.
(`shopt -s expand_aliases`)

[noissue]


bmbouter (approved):
Thanks!

dkliban (approved):

PR #701, opened 2022-10-04
Fix some tests by setting the pytest rootdir

mikedep333:
to /var/lib/pulp , where the pulp user can write.

[noissue]


mikedep333:
This actually didn't help change the location of ansible.cfg. It still gets written to /

PR #707, opened 2022-10-18
Pin azure-storage-blob until empty file azurite bug is released

daviddavis:
[noissue]


daviddavis:
I guess a question might be why not pin azure-blob-storage for all of Pulp? This problem is exclusive to Azurite (which returns a 206 instead of a 416) and azurite is only used by our CI. Normally, users will use Azure Blog Storage in production and I think they should use the latest version of azure-blob-storage.

fao89 (approved):

bmbouter:
Thank you!

Issue #709, opened 2022-10-24
static content is not present on the latest pulp_ci_centos images


mikedep333:
With this PR:
https://github.com/pulp/pulp-oci-images/pull/311

No longer does the pulpcore-api startup script collect static content.

Instead, the Containerfile that installs the plugins must collect static content.

So the plugin template's Containerfile must do that.

PR #710, opened 2022-10-24
Fix static content not existing in the images

mikedep333:
Previously, it was generated by the pulpcore-api startup script.

Now the Containerfile must collect static content.

fixes: #709


templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -27,7 +27,11 @@ RUN pip3 install \
 {{ " " }}"{{ item.source }}"
 {%- endfor %}
 
-RUN mkdir -p /etc/nginx/pulp/
+USER pulp:pulp
+RUN PULP_STATIC_ROOT=/var/lib/operator/static/ PULP_CONTENT_ORIGIN=localhost \
+       /usr/local/bin/pulpcore-manager collectstatic --clear --noinput --link
+USER root:root
```
bmbouter, line 33: Do we want to change back to `root:root`?

mikedep333: What are changing back to root:root is the context under which all subsequent commands like "RUN" in the Containerfile run as.

We want the rest of the Containerfile, including the symlinking, to do that.

fao89 (commented):
you need to add:
```python
STATIC_ROOT = "/var/lib/operator/static/"
```
here: https://github.com/pulp/plugin_template/blob/main/templates/github/.ci/ansible/settings.py.j2.copy

bmbouter (approved):
Thanks!

mikedep333:
> you need to add:
> 
> ```python
> STATIC_ROOT = "/var/lib/operator/static/"
> ```
> 
> here: https://github.com/pulp/plugin_template/blob/main/templates/github/.ci/ansible/settings.py.j2.copy

We dot not actually. It is set at the image layer so that single-container users do not have to set it:
https://github.com/pulp/pulp-oci-images/blob/latest/images/pulp_ci_centos/Containerfile#L12

fao89:
> > you need to add:
> > ```python
> > STATIC_ROOT = "/var/lib/operator/static/"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > here: https://github.com/pulp/plugin_template/blob/main/templates/github/.ci/ansible/settings.py.j2.copy
> 
> We do not actually. It is set at the image layer so that single-container users do not have to set it: https://github.com/pulp/pulp-oci-images/blob/latest/images/pulp_ci_centos/Containerfile#L12

I didn't notice that!

fao89 (approved):

PR #716, opened 2022-12-03
Installs functional test deps if pytest_args is specified.

dkliban:
Also renames TELEMETRY setting to ANALYTICS.

[noissue]

### DIFF
--- a/templates/github/.ci/ansible/settings.py.j2.copy
+++ b/templates/github/.ci/ansible/settings.py.j2.copy
@@ -8,7 +8,7 @@ TOKEN_SIGNATURE_ALGORITHM = "ES256"
 CACHE_ENABLED = True
 REDIS_HOST = "localhost"
 REDIS_PORT = 6379
-TELEMETRY = False
+ANALYTICS = False
 
 {% if api_root is defined %}
 API_ROOT = {{ api_root | repr }}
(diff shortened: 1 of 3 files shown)


templates/github/.ci/ansible/settings.py.j2.copy
```diff
@@ -8,7 +8,6 @@ TOKEN_SIGNATURE_ALGORITHM = "ES256"
 CACHE_ENABLED = True
 REDIS_HOST = "localhost"
 REDIS_PORT = 6379
-TELEMETRY = False
```
mdellweg, line 11: Don't remove that, replace it by ANALYTICS please.

mdellweg (changes requested):

templates/github/.ci/ansible/settings.py.j2.copy
dkliban: will do.

templates/github/.ci/ansible/settings.py.j2.copy
```diff
@@ -8,7 +8,7 @@ TOKEN_SIGNATURE_ALGORITHM = "ES256"
...
+ANALYTICS = False
```
mdellweg, line 11: Thanks. The reason is that the dev.analytics... will still happily sum up all systems, as we set the `..._MIN_AGE=0` there.

mdellweg (approved):

mdellweg:
I have no idea what the CI is doing on a monday again.

PR #720, opened 2023-01-04
Add GCP fake credentials

fao89:
https://django-storages.readthedocs.io/en/latest/backends/gcloud.html#authentication

Required PR: https://github.com/pulp/pulpcore/pull/3481

Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -121,6 +121,7 @@ services:
       - ../../../pulp-openapi-generator:/root/pulp-openapi-generator
     env:
       PULP_WORKERS: "4"
+      GOOGLE_APPLICATION_CREDENTIALS: "/etc/pulp/credentials.json"
 VARSYAML
 
 cat >> vars/main.yaml << VARSYAML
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -204,9 +204,17 @@ fi
 {%- endif %}
 
 {%- if test_gcp %}
-
+cat >> credentials.json << GCP_JSON
+{
```
ipanova, line 208: just to understand this better, where this is taken from?

fao89: from:
https://cloud.google.com/sdk/docs/install
https://cloud.google.com/docs/authentication/provide-credentials-adc#local-dev

ipanova: I pulled in these changes into my pr and the issue seems to persist https://github.com/pulp/pulp_container/pull/1196

fao89: it seems it needs more hacks:
https://github.com/fsouza/fake-gcs-server#using-with-signed-urls

ipanova:
@fao89 I don't think your changes are still enough, because pulp-container CI fails. Will you be able to test these changes on the example of pulp-container https://github.com/pulp/pulp_container/pull/1196

ipanova:
@pulp/deployments if someone can take this over and finalize, it  would be great. We have already merged the corresponding PR in core that states we GCP is supported https://github.com/pulp/pulpcore/pull/3481 hovewer we do not have a proof of a successful flow because the CI part was not done.

fao89:
actually, this PR is more related to @pulp/continuous-integration team

PR #725, opened 2023-01-23
Set an environment variable for test fixtures

lubosmj:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -141,6 +141,8 @@ if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   - name: pulp-fixtures\
     image: docker.io/pulp/pulp-fixtures:latest\
     env: {BASE_URL: "http://pulp-fixtures:8080"}' vars/main.yaml
+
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -141,6 +141,8 @@ if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   - name: pulp-fixtures\
     image: docker.io/pulp/pulp-fixtures:latest\
     env: {BASE_URL: "http://pulp-fixtures:8080"}' vars/main.yaml
+
+  export REMOTE_FIXTURES_ORIGIN="http://pulp-fixtures:8080"
```
mdellweg, line 145: So this is only in case `docker-fixtures` is set. So the `https://fixtures.pulpproject.org/` is the default value in the  corresponding pulpcore fixture, right?

mdellweg (approved):

templates/github/.github/workflows/scripts/install.sh.j2
lubosmj: Yes. Default value is `https://fixtures.pulpproject.org/`.

PR #726, opened 2023-02-16
Remove old redmine code

mdellweg:


plugin-template
```diff
@@ -30,8 +30,8 @@ DEFAULT_SETTINGS = {
     "coverage": False,
     "deploy_client_to_pypi": True,
     "deploy_client_to_rubygems": True,
-    "deploy_daily_client_to_pypi": True,
-    "deploy_daily_client_to_rubygems": True,
+    "deploy_daily_client_to_pypi": False,
+    "deploy_daily_client_to_rubygems": False,
```
mdellweg, line 34: I would like to remove the daily client pushing too. I cannot believe someone is using them. I still think we publish bindings with a release.

decko (approved):

PR #727, opened 2023-02-16
Remove Required-PR mechanism

mdellweg:
[noissue]

Experimentation: https://github.com/pulp/pulp_file/pull/897


lubosmj:
Pretty bold step. Do we think about any replacement for this? How will this change affect the current procedure for demonstrating the validity of added features in pulpcore? For instance, I used to implement a new feature in pulpcore and then write tests for it in pulp_file while referencing the commit with the feature from pulpcore.

mdellweg:
> Pretty bold step. Do we think about any replacement for this? How will this change affect the current procedure for demonstrating the validity of added features in pulpcore? For instance, I used to implement a new feature in pulpcore and then write tests for it in pulp_file while referencing the commit with the feature from pulpcore.

Thank you for asking! I think you can still add a line like this
`git+https://github.com/mdellweg/pulp-cli.git@squeezer_integration#egg=pulp-glue&subdirectory=pulp-glue`
to the requirements.txt.
It should allow the CI to run, yet fail the "ready-to-ship" test.

dkliban:
I had to add teh following diff to get the release pipeline to get past the 'build artifacts' stage:
``` diff
diff --git a/templates/github/.github/workflows/release.yml.j2 b/templates/github/.github/workflows/release.yml.j2
index 768fd63..c3c7c74 100644
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -47,9 +47,6 @@ jobs:
 
       {{ set_secrets() | indent(6) }}
 
-      - name: Verify upper bound requirements
-        run: python .ci/scripts/upper_bound.py
-
       - name: Create the release commit, tag it, create a post-release commit, and build plugin package
         run: python .github/workflows/scripts/release.py {{ "${{ github.event.inputs.release }}" }}
```

mdellweg:
@dkliban Thank you! I removed that check from the release workflow. There is no point i checking again if we checked that with every PR that constituted to this state.

dkliban (approved):
I tested these changes for both PRs and release workflow with pulp_maven. Thank you!

mdellweg:
I plan to merge on monday to get a head start for the automatic ci update on the weekend.

PR #732, opened 2023-03-24
Allow to run cli tests with plugin package 

mdellweg:


mdellweg:
@quba Might interest you too.

PR #733, opened 2023-03-24
Fix cli tests if none was found

mdellweg:
[noissue]


lubosmj (approved):

mdellweg:
Let's not merge this right. now. Maybe it is good to have an indicator that no tests were found when there should be some. We can just as well not activate the cli tests for those cases.

PR #734, opened 2023-03-24
Fix the search bar bug

lubosmj:
[noissue]

ref: https://github.com/pulp/pulpcore/pull/3668

### DIFF
--- a/templates/docs/docs/conf.py.j2
+++ b/templates/docs/docs/conf.py.j2
@@ -30,7 +30,9 @@ except ImportError:
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.extlinks']
+extensions = ['sphinx.ext.extlinks', 'sphinx.ext.autodoc', 'sphinx.ext.autosummary',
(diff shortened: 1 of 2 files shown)


templates/docs/docs/conf.py.j2
```diff
@@ -30,7 +30,9 @@ except ImportError:
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.extlinks']
+extensions = ['sphinx.ext.extlinks', 'sphinx.ext.autodoc', 'sphinx.ext.autosummary',
```
ggainey, line 33: Do these need to all be called out in doc_requirements as well?

ipanova:
@lubosmj these are templates. They will not be updated, they will only be generated once during plugin bootstrapping.

ipanova:
Also, i believe not every plugin uses `sphinx-rtd-theme`, or at least we're not enforcing that.

lubosmj:
Seems like there are more problems than with the original issue. You are both right about declaring unnecessary requirements.

If the template will not be updated, there is no need to continue working on this PR.

ggainey:
> If the template will not be updated, there is no need to continue working on this PR.

Aye, you're right. Apologies for suggesting you do something here for this - good thing we have @ipanova to keep us on track :)

PR #735, opened 2023-03-28
Fix availability of ci_requirements.txt in release

mdellweg:
[noissue]

### DIFF
--- a/.github/workflows/bootstrap.yaml
+++ b/.github/workflows/bootstrap.yaml
@@ -4,9 +4,6 @@ on:
   pull_request:
     branches:
       - '*'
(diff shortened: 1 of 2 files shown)


.github/workflows/bootstrap.yaml
```diff
@@ -4,9 +4,6 @@ on:
   pull_request:
     branches:
       - '*'
-  push:
-    branches:
-      - '*'
```
mdellweg, line 9: Not technically related, but these test runs did not even work, yet had absolutely no consequence.

lubosmj (approved):

PR #736, opened 2023-03-28
Fix client installations for older status endpoint

mdellweg:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -59,12 +59,29 @@ password password
 cmd_user_stdin_prefix bash -c "chmod 600 ~pulp/.netrc"
 
 cd ../pulp-openapi-generator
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -59,12 +59,29 @@ password password
...
 cd ../pulp-openapi-generator
-for item in $(echo "$REPORTED_STATUS" | jq -r '.versions[].package|sub("-"; "_")')
-do
-./generate.sh "${item}" python
-cmd_prefix pip3 install "/root/pulp-openapi-generator/${item}-client"
-sudo rm -rf "./${item}-client"
-done
+if [ "$(echo "$REPORTED_STATUS" | jq -r '.versions[0].package')" = "null" ]
+then
+  # We are on an old version of pulpcore without package in the status report
```
mdellweg, line 64: Legacy...

PR #738, opened 2023-03-29
Fix paths in taring the release working directory

mdellweg:
[noissue]

Obviously needs to be tested on a real release...


lubosmj (approved):
Let's give it a try.

PR #739, opened 2023-03-29
Fix copy of ci_requirements in container image

mdellweg:
[noissue]


ggainey (approved):
worked to get core/3.21.6 out

PR #744, opened 2023-04-21
Allow docs building without pulpcore

mdellweg:
[noissue]


lubosmj (approved):
Working: https://github.com/pulp/pulp_file/pull/952/files.

PR #747, opened 2023-05-05
Add concurrency statements to workflows

mdellweg:
[noissue]

This will attempt to cancel any running CI job on a PR in case the branch of the PR got updated. This should free us up some valuable CI time on a hot day.


ggainey:
I don't have enough context to know what we're doing here and why :(

ggainey (approved):

PR #749, opened 2023-05-11
Allow opentelemetry deps at beta

bmbouter:
[noissue]


ggainey (approved):
I'm ok with something that's "a little hackish" - it'll make us look hard at extending exceptions like this. Approved.

bmbouter:
@mdellweg please let me know what you think also. My goal was to address your concerns; hopefully I've done that, but let me know. Thank you!

mdellweg:
There was a check for the client package two lines down. I'd prefer it be specific to pulpcore. Also we agreed to add a large disclamer. I don't think this warrants for a followup.

bmbouter:
@mdellweg I'm not sure what the suggested changes are or I'd make them. Help me understand and I'm willing to do a little more.

PR #751, opened 2023-06-08
Remove tests on release

gerrod3:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -5,7 +5,6 @@
   setup_python,
   setup_ruby,
   install_httpie,
-  setup_env,
   display_logs,
   run_script,
   set_secrets,
@@ -22,7 +21,7 @@ on:
         required: true
       before_script:
         description: |
-          Bash code to run before script.sh is executed. This should only be used when re-running
+          Bash code to run before bindings and docs are built. This should only be used when re-running
           a workflow to correct some aspect of the docs. e.g.: git checkout origin/3.14 CHANGES.rst
         required: false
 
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/release.yml.j2
```diff
@@ -115,8 +91,8 @@ jobs:
 
       {{ install_httpie() | indent(6) }}
 
-      {{ setup_env() | indent(6) }}
-
+      # Building the bindings and docs requires accessing the OpenAPI specs endpoint, so we need to
+      # setup the Pulp instance.
```
mdellweg, line 95: Actually, the content-app container should be sufficient... :thinking:

templates/github/.github/workflows/release.yml.j2
```diff
@@ -127,56 +103,55 @@ jobs:
...
+      {%- endif %}
+
+      - name: Build docs
+        run: |
+          pip install towncrier==19.9.0
+          towncrier --yes --version 4.0.0.ci
+          make -C docs/ PULP_URL="{{ pulp_scheme }}://pulp" diagrams html
+          tar -cvf docs/docs.tar docs/_build
 
-      {%- if docs_test %}
```
mdellweg, line 165: This is a plugin-template switch. I think for plugins that do not have/build docs we need to keep this, or consolidate with `publish_docs_to_pulpprojectdotorg`.

templates/github/.github/workflows/release.yml.j2
gerrod3: Do you mean the pulp-api app container?

mdellweg: Oh, sorry, unrelated. I was just thinking what the minimal requirement were to get the api.json. A container allowing to run pulpcore-manager should be sufficient, but in the ci we always fire up the complete single container. So please ignore!

mdellweg (approved):

Issue #754, opened 2023-06-15
Remove entries from CHANGES directory when bumping version


dkliban:
The PR that is created to bump the development version on main branch after a Y branch was created should delete all the entries in the CHANGES directory. 

Also need to stop opening changelog PRs after a release.

PR #755, opened 2023-06-15
Deletes entries in CHANGES directory after creatinga new Y branch

dkliban:
Also stops opening changelog PRs after a release.

fixes: #754

### DIFF
--- a/templates/github/.github/workflows/create-branch.yml.j2
+++ b/templates/github/.github/workflows/create-branch.yml.j2
@@ -52,7 +52,11 @@ jobs:
           git checkout {{ plugin_default_branch }}
           bump2version --no-commit minor
 
-      - name: Make a PR with version bump
+      - name: Remove entries from CHANGES directory
+        run: |
+          find CHANGES -type f -regex ".*\.\(bugfix\|doc\|feature\|misc\|deprecation\|removal\)" -exec git rm {} +
+
+      - name: Make a PR with version bump and without CHANGES/*
         uses: peter-evans/create-pull-request@v4
         with:
           token: {{ '${{ secrets.RELEASE_TOKEN }}' }}
--- a/templates/github/.github/workflows/release.yml.j2
+++ b/templates/github/.github/workflows/release.yml.j2
@@ -279,25 +279,3 @@ jobs:
 
       - name: Create release on GitHub
         run: bash .github/workflows/scripts/create_release_from_tag.sh {{ "${{ github.event.inputs.release }}" }}
-
-      - name: Cleanup repository before making changelog PR
-        run: rm -rf .lock generation {{ plugin_snake }}_client* *-client.tar {{ plugin_name }}.tar todo web *docs.tar
-
-      - name: Stage changelog for {{ plugin_default_branch }} branch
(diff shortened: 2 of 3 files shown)


templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -52,7 +52,11 @@ jobs:
           git checkout {{ plugin_default_branch }}
           bump2version --no-commit minor
 
-      - name: Make a PR with version bump
+      - name: Remove entries from CHANGES directory
+        run: |
+          git rm CHANGES/*
```
gerrod3, line 57: I think this also deletes the dot files and the CHANGES folder itself.

gerrod3: Have you tested the command on your machine? On mine in an 'empty' plugin CHANGES directory it deleted both dot files and the directory itself. Maybe this isn't a fair comparison if we expect there to always be changelog entries present.

```
[gerrod@localhost pulp_python]$ git rm CHANGES/*
rm 'CHANGES/.TEMPLATE.rst'
rm 'CHANGES/.gitignore'
[gerrod@localhost pulp_python]$ ls -la CHANGES/
ls: cannot access 'CHANGES/': No such file or directory
```

dkliban: Your shell has globstar enabled. I'll improve the patch to be more specific and only remove entries.

templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -52,7 +52,11 @@ jobs:
           git checkout {{ plugin_default_branch }}
           bump2version --no-commit minor
 
-      - name: Make a PR with version bump
+      - name: Remove entries from CHANGES directory
+        run: |
+          find CHANGES -type f -regex ".*\.\(bugfix\|doc\|feature\|misc\)" -exec git rm {} +
```
gerrod3, line 57: There's two more: removal & deprecation. https://github.com/pulp/pulpcore/blob/main/pyproject.toml#L20-L50

gerrod3 (changes requested):

templates/github/.github/workflows/create-branch.yml.j2
dkliban: Great eye! I added both!

templates/github/.github/workflows/release.yml.j2
```diff
@@ -285,19 +285,3 @@ jobs:
 
       - name: Stage changelog for {{ plugin_default_branch }} branch
         run: python .github/workflows/scripts/stage-changelog-for-default-branch.py {{ "${{ github.event.inputs.release }}" }}
```
gerrod3, line 287: I think this step, the one above it (Cleanup repository before making changelog PR) and the workflow script (stage-changelog-for-default-branch.py) can all be removed as well.

gerrod3 (approved):
@dkliban Nice! One less PR to manually review.

templates/github/.github/workflows/create-branch.yml.j2
mdellweg: Actually, there is a default set or it is configured in pyproject.yaml. How much dynamic do we want to make it?

mdellweg (approved):

PR #757, opened 2023-06-27
Fix docs build for pulpcore during release

gerrod3:
These two env vars need to be set for pulpcore docs building outside of the pulp container, since pulpcore requires django to build some plugin api docs.


jctanner (approved):
:+1:

ggainey (approved):
lgtm

PR #763, opened 2023-07-05
Stop calling towncrier with docs for release

mdellweg:
The commit containing that change has already happened at this stage.

[noissue]


ggainey (changes requested):
If we're not going to call it, we shouldn't install towncrier at line#43

mdellweg:
> If we're not going to call it, we shouldn't install towncrier at line#43

Line 43 is part of the other job (running in a different vm) that is calling towncrier as part of the release.py script in line 50.

ggainey (approved):

PR #765, opened 2023-07-11
Look for template_gitref files instead of using hardcoded paths

dkliban:


gerrod3 (approved):

templates/github/.github/workflows/scripts/update_ci.sh.j2
```diff
@@ -15,7 +15,7 @@ popd
 # Check if only gitref file has changed, so no effect on CI workflows.
 if [ "$(git diff --name-only | grep -v "template_gitref")" ]; then
   echo "No changes detected."
-  git restore ".github/template_gitref" "docs/template_gitref"
+  git ls-files | grep template_gitref | xargs git restore
```
mdellweg, line 18: I wonder what went wrong here...

PR #766, opened 2023-07-11
Moves update_ci.sh to plugin_template repo

dkliban:
This also adds a new script for reading values from template_config.yml. The update_ci.sh script uses this
script to read the template_config.yml on the branch that is being updated.
    
This also fixes how changes are detected after plugin_template is applied to a branch.
    
[noissue]


mdellweg (changes requested):
I don't think this is right. And if it is, we should move that script to the plugin_template repository.

templates/github/.github/workflows/update_ci.yml.j2
```diff
@@ -44,6 +44,11 @@ jobs:
           ref: '{{ branch }}'
           fetch-depth: 0
 
+      - name: Checkout update_ci.sh from main
+        working-directory: {{ plugin_name }}
+        run: |
+          git checkout main -- .github/workflows/scripts/update_ci.sh
```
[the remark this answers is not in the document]
dkliban: This is needed because the branch to which the plugin template is being applied to is checked out before the update_ci.sh script is run. We are now in a situation where there is a bug in the update_ci.sh on the 3.28 branch and even though we fixed it on the main branch, the script is executed from the 3.28 branch.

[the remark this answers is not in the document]
dkliban: I am thinking that instead of checking out the file from main branch, we should just make a copy of it on the main branch and then use that copy of it.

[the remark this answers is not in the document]
dkliban: I now understand what you are saying. Yes, let's move the script to the plugin_template repository and call it from there!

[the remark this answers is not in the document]
dkliban: @gerrod3 I implemented the changes suggested by @mdellweg! Thanks for the suggestion @mdellweg

gerrod3 (approved):
Looks correct to me.

scripts/update_ci.sh
```diff
@@ -9,18 +8,21 @@ if [ ! -f "template_config.yml" ]; then
 fi
 
 pushd ../plugin_template
-./plugin-template --github {% if ci_update_docs -%} --docs {% endif -%} {{ plugin_name }}
+# Pass on all but the last argument (which is the noissue_marker)
+args="${@:1:$#-1}"
```
mdellweg, line 12: Can't we just read the noissue marker from the template_config?

templates/github/.github/workflows/update_ci.yml.j2
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: Run update
         working-directory: {{ plugin_name }}
         run: |
-          .github/workflows/scripts/update_ci.sh
+          ../plugin_template/scripts/update_ci.sh {% if ci_update_docs -%} --docs {% endif -%} {{ plugin_name }} {{ noissue_marker | default('[noissue]') }}
```
mdellweg, line 50: Yes, looks right to me. This now even covers the case where the update script on main would have been broken.
Can you add the old one to the deprecated_files please?

mdellweg (changes requested):

scripts/update_ci.sh
dkliban: We probably could, but I did not want to try to parse template_config here.

mdellweg: I think we need to, as it could be different on different branches.

scripts/update_ci.sh
```diff
@@ -0,0 +1,39 @@
...
+set -eu
+
+if [ ! -f "template_config.yml" ]; then
+  echo "No template_config.yml detected."
+  exit 1
+fi
+
+plugin_name = python ../plugin_template/scripts/get_template_config_value.py plugin_name
+ci_update_docs = python ../plugin_template/scripts/get_template_config_value.py ci_update_docs
+noissue_marker = python ../plugin_template/scripts/get_template_config_value.py noissue_marker
```
mdellweg, line 12: Are you sure this works in bash? I thought i'd be:
```suggestion
plugin_name="$(python ../plugin_template/scripts/get_template_config_value.py plugin_name)"
ci_update_docs="$(python ../plugin_template/scripts/get_template_config_value.py ci_update_docs)"
noissue_marker="$(python ../plugin_template/scripts/get_template_config_value.py noissue_marker)"
```

scripts/update_ci.sh
```diff
@@ -0,0 +1,39 @@
...
+
+if [[ -z "$noissue_marker" ]]; then
+  noissue_marker = "[noissue]"
```
mdellweg, line 15: Again my bash experience:
```suggestion
  noissue_marker="[noissue]"
```

scripts/update_ci.sh
```diff
@@ -0,0 +1,39 @@
...
+
+if [[ ci_update_docs == "True"]]; then
+  docs = "--docs"
```
mdellweg, line 18:
```suggestion
  docs="--docs"
```

scripts/update_ci.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+  docs = ""
```
mdellweg, line 20:
```suggestion
  docs=""
```

mdellweg (approved):

PR #767, opened 2023-07-12
Add ability to update ci_update_branches from create-branch workflow

gerrod3:
~~Decided to add this functionality to the plugin-template script and have the create-branch workflow call with the new options if specified at dispatch time. Here is how the new dispatch screen looks:~~

Screenshot from 2023-07-12 01-31-28

Plugins have different behavior from pulpcore, some don't even set this variable, so I added three different options to cover each scenario.

*EDIT*
Changed it to a new template_config variable: `ci_update_release_behavior`.

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -30,6 +30,7 @@ DEFAULT_SETTINGS = {
     "ci_trigger": "{pull_request: {branches: ['*']}}",
     "ci_update_branches": [],
     "ci_update_docs": False,
+    "ci_update_release_behavior": None,
     "core_import_allowed": [],
     "deploy_client_to_pypi": True,
     "deploy_client_to_rubygems": True,
@@ -218,6 +219,19 @@ def main():
                         """
         ),
     )
+    parser.add_argument(
+        "--add-version-ci-update-branches",
+        metavar="VERSION_BRANCH",
+        action="store",
+        help=textwrap.dedent(
+            """\
+                            Add specified version to the ci_update_branches before templating.
+                            This new value will replace the previous latest version if
+                            `ci_update_release_behavior` is set to 'replace-previous-version'.
+
+                        """
+        ),
+    )
     args = parser.parse_args()
 
     here = os.path.dirname(os.path.abspath(__file__))
@@ -270,6 +284,26 @@ def main():
                 config = generate_config(args.plugin_name, args.plugin_app_label)
         else:
             return 2
+
+    valid_release_behavior = [None, "append-version", "replace-previous-version"]
+    if config["ci_update_release_behavior"] not in valid_release_behavior:
+        valid_release_behavior[0] = "null"
+        print("ci_update_release_behavior is not of a valid value: ", valid_release_behavior)
+        return 2
+
+    if new_version_branch := args.add_version_ci_update_branches:
+        write_new_config = True
+        if config["ci_update_release_behavior"] == "replace-previous-version":
+            major_v, minor_v = new_version_branch.split(".", maxsplit=1)
+            minor_v = int(minor_v) - 1
+            try:
+                index = config["ci_update_branches"].index(f"{major_v}.{minor_v}")
+            except ValueError:
+                config["ci_update_branches"].append(new_version_branch)
+            else:
+                config["ci_update_branches"][index] = new_version_branch
+        else:
+            config["ci_update_branches"].append(new_version_branch)
 
     # Config key is used by the template_config.yml.j2 template to dump
     # the config. (note: uses .copy() to avoid a self reference)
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -13,6 +13,15 @@ on:
...
         required: true
+      update_ci_branches:
+        description: "Behavior for updating ci_update_branches"
+        required: true
+        type: choice
+        default: "none"
+        options:
+          - "append-version"
+          - "replace-previous-version"
+          - "none"
```
mdellweg, line 24: Interesting idea, but i wonder if we should make this choice in the template-config. Giving the user choice usually leads to confusion.
I suspect we will hear: "Did you not select this option? We always do. Now I need to do it by hand."

gerrod3: Like another variable to decide the behavior? I was also thinking of just having the default be different for pulpcore, have it be set to replace-previous-version instead of none since this is the behavior we've decided on for our release policy.

gerrod3: What do you suggest then? Switch it to a template-config variable?

dkliban:
Let's make this a template_config variable.

plugin-template
```diff
@@ -217,6 +218,28 @@ def main():
...
+        "--replace-previous-version-ci-update-branches",
+        action="store_true",
+        help=textwrap.dedent(
+            """\
+                            Used with --add-version-ci-update-branches, replaces the previous
+                            version in ci_update_branches with the new version.
+
+                        """
+        ),
+    )
```
mdellweg, line 242: If i'm not mistaken, the script should be able defer this option from the template_config.

plugin-template
```diff
@@ -217,6 +218,28 @@ def main():
                         """
         ),
     )
+    parser.add_argument(
+        "--add-version-ci-update-branches",
```
mdellweg, line 222: Not sure, the other one is actually more descriptive...
```suggestion
        "--add-ci-branch",
```

So am i understanding correctly that this is another option like `--github` or `--docs` that specifies what parts of the plugin to change? I guess, it implies `--github`?

plugin-template
```diff
@@ -270,6 +293,26 @@ def main():
...
+            major_v, minor_v = new_version_branch.split(".", maxsplit=1)
+            minor_v = int(minor_v) - 1
+            try:
+                index = config["ci_update_branches"].index(f"{major_v}.{minor_v}")
+            except ValueError:
+                config["ci_update_branches"].append(new_version_branch)
+            else:
+                config["ci_update_branches"][index] = new_version_branch
+        else:
+            config["ci_update_branches"].append(new_version_branch)
```
mdellweg, line 314: So we are assuming here, that a new major version will never replace the last version.
This sounds fine to me.

plugin-template
```diff
@@ -270,6 +293,26 @@ def main():
...
 
+    valid_release_behavior = [None, "append-version", "replace-previous-version"]
+    if config["ci_update_release_behavior"] not in valid_release_behavior:
+        valid_release_behavior[0] = "null"
+        print("ci_update_release_behavior is not of a valid value: ", valid_release_behavior)
+        return 2
+
+    if new_version_branch := args.add_version_ci_update_branches:
+        write_new_config = True
+        if args.replace_previous_version_ci_update_branches:
```
mdellweg, line 304: This could be read from the config directly.

templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -55,7 +55,24 @@ jobs:
...
+          path: plugin_template
+          fetch-depth: 0
 
+      - name: Update CI branches in template_config
+        working-directory: plugin_template
+        run: |
+          ARGS="--add-version-ci-update-branches {{ "${{ inputs.name }}" }}"
+          {%- if ci_update_release_behavior == 'replace-previous-version' %}
+          ARGS="$ARGS --replace-previous-version-ci-update-branches"
+          {%- endif %}
```
mdellweg, line 72: See the suggestion to read this from the config in plugin-template.

plugin-template
gerrod3: The option only modifies the variable in the config file, you have to use `--github` in order to update the templated files.

mdellweg: So yes in a way that it implies you also want to specify `--github`. XD

plugin-template
mdellweg: Replacing only works for 2.1 -> 2.2. If we released 2.1 -> 3.0 we would still keep 2.1 in the list. And that's probably fine because OTOH it's very rare, OTOH it's probably wise to support the last y of the previous x for a bit longer.

mdellweg (approved):

PR #769, opened 2023-07-17
Improve collect changes

mdellweg:
[noissue]


.ci/bootstrap_catdog.sh
```diff
@@ -9,7 +9,7 @@ then
 else
     COMMIT_MSG="Initial commit\n\n[noissue]"
 fi
-echo $COMMIT_MSG
+echo "$COMMIT_MSG"
```
ggainey, line 12: If we want to be *really* correct, bash-params should be protected via ${PARAM}

ggainey (changes requested):

.ci/bootstrap_catdog.sh
mdellweg: I *think* that is optional. This change is what made shellcheck silent. ;)
```suggestion
echo "${COMMIT_MSG}"
```

ggainey (approved):

PR #781, opened 2023-09-11
Changes where plugin_template is cloned to

dkliban:
This ensures that the plugin-template command looks for the template_config.yaml in the right place.

[noissue]

### DIFF
--- a/templates/github/.github/workflows/create-branch.yml.j2
+++ b/templates/github/.github/workflows/create-branch.yml.j2
@@ -25,7 +25,10 @@ jobs:
       fail-fast: false
 
     steps:
-      {{ checkout() | indent(6) }}
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          path: {{ plugin_name }}
 
       {{ setup_python(pyversion="3.8") | indent(6) }}
 


templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -60,11 +60,11 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: pulp/plugin_template
-          path: plugin_template
+          path: ../plugin_template
```
mdellweg, line 63: Checkout will not let you.
I know this is crazy, but you may need to move both checkouts one directory down.

mdellweg: Instead you need to add `path: {{ plugin_name }}` to the other checkout.
And now we need to carefully look that all the other scripts still run properly.

mdellweg (changes requested):

templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -25,7 +25,10 @@ jobs:
       fail-fast: false
 
     steps:
-      {{ checkout() | indent(6) }}
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          path: {{ plugin_name }}
```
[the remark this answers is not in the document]
mdellweg: create-pull-request has a `path` argument.

https://github.com/peter-evans/create-pull-request#action-inputs

mdellweg:
Even if that makes things more complex, can we still rearrange the workflow so that we make all changes locally and then first create the PR for the main branch and then call `git push 3.yy`?
That way we have almost no window where the workflow can fail, but also have had side effects.

PR #787, opened 2023-10-04
Change installation of storage backend to use pulpcore's pip extra install

gerrod3:
[noissue]


mdellweg (approved):
This seems to be an improvement in any case.

PR #790, opened 2023-11-03
Remove update_github step from release pipeline

mdellweg:
[noissue]


mdellweg:
Please look if the ci is only read for the multiple commits in this PR.

dkliban (approved):

mdellweg:
Here's a question to answer. Given we have a workflow triggered by a tag. Will it run with the workflow file of the tag or the one on the main branch?

templates/github/.github/workflows/scripts/release.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Release ${NEW_VERSION}"
+
+if ! [[ "${NEW_VERSION}" == "${BRANCH}"* ]]
+then
+  echo ERROR: Version does not match release branch
+  exit 1
+fi
+
+towncrier build --yes --version "${NEW_VERSION}"
+bumpversion release --commit --message "Release {new_version}" --tag --tag-name "{new_version}" --tag-message "Release {new_version}" --allow-dirty
```
dkliban, line 23: bump2version?

mdellweg: Should both work. But the readme is built on bump2version. We should use that.

dkliban (approved):

templates/github/.github/workflows/scripts/release.sh
ipanova: can we use bump2version everywhere, like we started on line 13 let's continue with that

ipanova (approved):

PR #791, opened 2023-11-03
Fix release script

mdellweg:
[noissue]


ggainey (approved):

mdellweg:
This now fixes pulp_file version not being properly updated.

dkliban (approved):
Let's try this out.

PR #794, opened 2023-11-13
Automatic backport label creation kebab case support

### LINKED ISSUE #788: update labels is using plugin_snake
This does not produce propoer repository names.

MichalPysik:
Github workflow automatically creating backport labels assumed all plugins' repositories use snake_case. This has now been fixed by introducing support for kebab-case named repositories too.

closes #788


mdellweg (changes requested):
We are in the plugin template here, we can **know** the right plugin name. No guesswork please.

mdellweg (approved):

PR #801, opened 2023-11-16
Cleanups

mdellweg:


dkliban (approved):

mdellweg:
I need to check whether the changed paths interfere with the release wf.

mdellweg:
@dkliban I removed the relocated checkouts, because it seemed to dangerous now.

dkliban (approved):

PR #802, opened 2023-11-20
Remove option for single commit check

mdellweg:
The new version of this check is no longer intimidating, and switching off this option leads to an invalid workflow file.

[noissue]


lubosmj (approved):
Nice!

PR #803, opened 2023-11-21
Use a better key for the static PR check

mdellweg:
Since this workflowruns against main, the ref_name is not suitable for a cancellation condition.

[noissue]


mdellweg:
I wish i could test this somewhere.

lubosmj (approved):

PR #805, opened 2023-11-22
Remove workaround not publishing pulp-file client

mdellweg:
Since we can now publish clients for multiple plugins in one repository, this workaround is not needed anymore. Also it turned out to be evil on older pulpcore branches by installing the wrong version of the bindings.

[noissue]


lubosmj (approved):
Squashing commits would be favourable.

mdellweg:
I'd like to be able to revert them individually.
They are really solving two different cases.

PR #811, opened 2023-11-28
Fix deprecation warnings

mdellweg:
This is yet another legacy of the pulp_file merge.

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -3,15 +3,10 @@
   set_env_vars,
   checkout,
   setup_python,
-  install_httpie,
-  setup_env,
-  display_logs,
-  run_script,
-  set_secrets,
   install_python_deps,
 with context %}
 ---
-name: {{ plugin_app_label | camel }} CI
+name: "{{ plugin_app_label | camel }} CI"
 on: {{ ci_trigger | default("{pull_request: {branches: ['*']}}") }}
 
 concurrency:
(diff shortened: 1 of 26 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -27,46 +22,46 @@ jobs:
...
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-latest"
     steps:
       {{ checkout(depth=0, path=plugin_name) | indent(6) }}
       {{ setup_python() | indent(6) }}
       {%- if check_commit_message %}
-      - name: Install requirements
-        run: pip3 install github
-      - name: Check commit message
+      {{ install_python_deps(["requests", "pygithub"]) | indent(6) }}
```
lubosmj, line 31: We install missing packages in the scripts themselves; the question is, do we need this^?

https://github.com/pulp/plugin_template/blob/f2059547f44036fef490e00863e058ce85881076/templates/github/.github/workflows/scripts/check_commit.sh.j2#L13

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -74,7 +73,7 @@ jobs:
 
       {{ setup_python() | indent(6) }}
 
-      {{ install_python_deps("requests") | indent(6) }}
+      {{ install_python_deps(["requests"]) | indent(6) }}
```
lubosmj, line 76: I would also install this requirement inside the caller ".github/workflows/scripts/publish_docs.sh". Do you think it is worth it to revisit all installing dependencies and move them to one place?

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -63,31 +63,15 @@ jobs:
...
-          {%- if test_azure %}
-          echo "{{ "${{ needs.test.outputs.deprecations-azure }}" }}" | base64 -d
-          {%- endif %}
-          {%- if test_s3 %}
-          echo "{{ "${{ needs.test.outputs.deprecations-s3 }}" }}" | base64 -d
-          {%- endif %}
-          {%- if test_lowerbounds %}
-          echo "{{ "${{ needs.test.outputs.deprecations-lowerbounds }}" }}" | base64 -d
-          {%- endif %}
+          cat deprecations-*.txt | sort -u
```
lubosmj, line 72: Do we miss "{{ plugin_name }}" in the path?

templates/github/.github/workflows/nightly.yml.j2
mdellweg: I intended to stop installing anything from scripts, because i consider that to be dangerous if you try to run any of these locally.

templates/github/.github/workflows/ci.yml.j2
mdellweg: At this point no. As we do no checkout, we don't have the directory "{{ plugin_name }}" to do that. Do you think we should create it for consistency?

lubosmj: I would say yes.

templates/github/.github/workflows/nightly.yml.j2
lubosmj: Gotcha, then it makes sense to remove pip calls from the scripts, in another PR.

templates/github/.github/workflows/ci.yml.j2
mdellweg: I thought i removed that.

templates/github/.github/workflows/nightly.yml.j2
mdellweg: And i dind't even check with "git grep" ... XD

templates/github/.github/workflows/ci.yml.j2
mdellweg: Actually, i did.

lubosmj (approved):

PR #813, opened 2023-11-29
Fix update ci workflow

mdellweg:
We actually need the git history in order to use git describe. Since this change prevented further CI updates, the call to plugin template has been made resilient to this failure.

[noissue]

### DIFF
--- a/templates/macros.j2
+++ b/templates/macros.j2
@@ -51,10 +51,10 @@ GITHUB_CONTEXT: "{{ '${{ github.event.pull_request.commits_url }}' }}"
 - name: "Install python dependencies"
   run: |
     echo ::group::PYDEPS
-    pip install {{ deps | join(" ") }}
-    {%- if 'httpie' in deps %}
+    pip install {{ deps | map("shquote") | join(" ") }}
+    §= if 'httpie' in deps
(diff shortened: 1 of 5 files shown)


templates/macros.j2
```diff
@@ -51,10 +51,10 @@ GITHUB_CONTEXT: "{{ '${{ github.event.pull_request.commits_url }}' }}"
 - name: "Install python dependencies"
   run: |
     echo ::group::PYDEPS
-    pip install {{ deps | join(" ") }}
-    {%- if 'httpie' in deps %}
+    pip install {{ deps | map("shquote") | join(" ") }}
+    §= if 'httpie' in deps
```
dkliban, line 55: I've never seen § used in templates.

mdellweg: https://jinja.palletsprojects.com/en/3.0.x/templates/#line-statements

I thought using "#" has some detrimental effects when heavily templating shell scripts. OTOH I believe using line statements will reduce time we need to think about these whitespace eating `-` a lot.

dkliban: Thank you! neat feature

dkliban (approved):

PR #822, opened 2023-12-06
Ensure correct dependencies are installed for lowerbounds and cloud storage tests

gerrod3:
Using MinIO for our S3 test backend requires us to use this patch to avoid timeouts when uploading 0-byte files.

[noissue]

### DIFF
--- a/templates/github/.ci/ansible/Containerfile.j2.copy
+++ b/templates/github/.ci/ansible/Containerfile.j2.copy
@@ -2,27 +2,24 @@ FROM {{ ci_base | default("ghcr.io/pulp/pulp-ci-centos:" + pulp_container_tag) }
 
 # Add source directories to container
 {% for item in plugins %}
-{% if item.source.startswith("./") or item.ci_requirements | default(false) %}
 ADD ./{{ item.name }} ./{{ item.name }}
-{% endif %}
 {% endfor %}
 
 # Install python packages
+# S3 botocore needs to be patched to handle responses from minio during 0-byte uploads
+# Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
 RUN pip3 install
+{%- if s3_test | default(false) -%}
+{{ " " }}git+https://github.com/fabricio-aguiar/botocore.git@fix-100-continue
+{%- endif -%}
 {%- for item in plugins -%}
 {%- if item.name == "pulp-certguard" -%}
 {{ " " }}python-dateutil rhsm
 {%- endif -%}
 {{ " " }}{{ item.source }}
-{%- if item.name == "pulpcore" -%}
-{%- if s3_test | default(false) -%}
-[s3]
-{%- elif azure_test | default(false) -%}
-[azure]
-{%- elif gcp_test | default(false) -%}
-[google]
-{%- endif -%}
+{%- if item.lowerbounds | default(false) -%}
+{{ " " }}-c ./{{ item.name }}/lowerbounds_constraints.txt
 {%- endif -%}
 {%- if item.ci_requirements | default(false) -%}
 {{ " " }}-r ./{{ item.name }}/ci_requirements.txt
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -61,8 +61,7 @@ then
 fi
 
 if [[ "$TEST" = "lowerbounds" ]]; then
-  python3 .ci/scripts/calc_deps_lowerbounds.py > lowerbounds_requirements.txt
-  mv lowerbounds_requirements.txt requirements.txt
+  python3 .ci/scripts/calc_deps_lowerbounds.py > lowerbounds_constraints.txt
 fi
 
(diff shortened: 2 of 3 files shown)


templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -17,7 +19,7 @@ RUN pip3 install
 {{ " " }}{{ item.source }}
 {%- if item.name == "pulpcore" -%}
 {%- if s3_test | default(false) -%}
-[s3]
+[s3] git+https://github.com/fabricio-aguiar/botocore.git@fix-100-continue
```
mdellweg, line 22: I wonder if we needed a rebased version of this fix. Also how do we handle this in production?
Do we need to vendor botocore?

gerrod3: It only affects MinIO, so if users are using that as their backend they will run into a timeout issue after a 0-byte file is uploaded. Luckily with Pulp's artifact system this will only occur once as the second time a 0-byte file is downloaded/uploaded the artifact will be re-used. Also, looking at the Deb tests that trigger this timeout it doesn't actually fail the sync since we have retries on downloads after the timeout. It would probably be good to use a rebased version of this fix.

Also, I've come to realize this bit of code for backend specific installs only works in the pulpcore repository since the `plugins` list will usually only have the plugin it's testing.

Second thing I've realized is that our lowerbounds test runner is no longer doing anything with the addition of the build step. We modify the `requirements.txt` after the build step has already ran. We probably need to use the output of `.ci/scripts/calc_deps_lowerbounds.py` as a constraint file for the pip install. I'll see if I can modify this PR to include that.

gerrod3:
I realized my earlier change to ensure django-storage swas installing from the pulpcore requirements was actually only occurring in the pulpcore repository. It just happens that we install django-storages[boto3,azure] in the base pulp image (https://github.com/pulp/pulp-oci-images/blob/latest/images/Containerfile.core.base#L79), so we didn't see any errors in the plugin repositories. Also, lowerbounds was no longer installing from the recalculated requirements.txt since we switched to the build step, so I fixed that by passing it in as a constraint file.

pulpcore test pr: https://github.com/pulp/pulpcore/pull/4849
pulp_python test pr: https://github.com/pulp/pulp_python/pull/638
Check out the pip list in the install step for lowerbounds and s3/azure runners, they are now correct.

templates/github/.ci/ansible/Containerfile.j2.copy
mdellweg: I think, i get what you say. We only pass the built whl to pip inside the container forgetting about the just adapted requirements.txt. I'll find a solution for that.

templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -62,7 +62,6 @@ fi
 
 if [[ "$TEST" = "lowerbounds" ]]; then
   python3 .ci/scripts/calc_deps_lowerbounds.py > lowerbounds_requirements.txt
-  mv lowerbounds_requirements.txt requirements.txt
```
mdellweg, line 65: I love this part of the change!
Shall we name the file `lowerbounds_contstraints.txt` instead?

templates/github/.ci/ansible/Containerfile.j2.copy
```diff
@@ -2,27 +2,29 @@ FROM {{ ci_base | default("ghcr.io/pulp/pulp-ci-centos:" + pulp_container_tag) }
...
+# S3 botocore needs to be patched to handle responses from minio during 0-byte uploads
+# Hacking botocore (https://github.com/boto/botocore/pull/1990)
 
 RUN pip3 install
+{%- if s3_test | default(false) -%}
+{{ " " }}git+https://github.com/fabricio-aguiar/botocore.git@fix-100-continue
+{%- endif -%}
 {%- for item in plugins -%}
 {%- if item.name == "pulp-certguard" -%}
 {{ " " }}python-dateutil rhsm
```
mdellweg, line 20: Can we move that into certguards ci-requirements?

mdellweg: Ooof :facepalm:.

Anyway:
https://github.com/pulp/pulp-certguard/pull/331

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -52,6 +58,34 @@ if [[ -f ../../ci_requirements.txt ]]; then
     ci_requirements: true
 VARSYAML
 fi
+{%- if test_s3 %}
+if [ "$TEST" = "s3" ]; then
+  cat >> vars/main.yaml << VARSYAML
+    pulpcore: "${PULPCORE_PREFIX}[s3]"
```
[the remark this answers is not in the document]
gerrod3: I refactored this bit so that the generated `install.sh` is easier to read, see https://github.com/pulp/pulp_python/pull/638/files#diff-02059c243ab1ae5faa6a95ec965f0b351e0f6c96e5452c90881b6b7e3ee11618R40-R46

PR #827, opened 2024-01-08
Use PyPI API token instead of username + password

dralley:


templates/github/.github/workflows/scripts/publish_plugin_pypi.sh.j2
```diff
@@ -22,7 +22,7 @@ then
   exit
 fi
 
-twine upload -u pulp -p "$PYPI_PASSWORD" \
+twine upload -u __token__ -p "$PYPI_API_TOKEN" \
```
mdellweg, line 25: I take it, you can still overwrite the token variable to serve another pypi account...

mdellweg (approved):

PR #829, opened 2024-01-10
Generate bindings from app label

mdellweg:
Starting with pulpcore 3.43 the package in the status api represents the proper python package for pulp_file (now being pulpcore). Therefore we can no longer depend on this to infer the name for the bindings package.

[noissue]


mdellweg:
Do not merge this before https://github.com/pulp/pulpcore/pull/4729 .

PR #839, opened 2024-01-30
Add a migration to add a version pin to pytest

mdellweg:
[noissue]

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -15,6 +15,7 @@ import yaml
 from pathlib import Path
 
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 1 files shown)


plugin-template
```diff
@@ -329,11 +330,44 @@ def main():
...
+        for line in lines:
+            try:
+                req = Requirement(line)
+            except ValueError:
+                result.append(line)
+                continue
+            if req.name == "pytest":
+                found = True
+                if not req.specifier:
+                    req.specifier = "<8"
```
ggainey, line 357: :+1:

ggainey (approved):

Issue #843, opened 2024-02-10
Have the ci_base image be configurable


gerrod3:
Currently you can only change the base image used to build the if you put `CI Base Image:` in your commit message (https://github.com/pulp/plugin_template/blob/0528507ec6678f7a5aaf73cb02ad0339a50c6f91/templates/github/.github/workflows/scripts/before_install.sh.j2#L48-L51). If not specified it will use the default image, pulp-ci-centos (https://github.com/pulp/plugin_template/blob/0528507ec6678f7a5aaf73cb02ad0339a50c6f91/templates/github/.ci/ansible/Containerfile.j2.copy#L1). With the release of the new centos9 images (https://github.com/pulp/pulp-oci-images/pull/592) with a new naming scheme we need to be able to change this default value for new branches, while keeping the old branches on the previous.

PR #844, opened 2024-02-12
Allow ci_base image to be configurable

gerrod3:
This will allow us to switch our latest branches to the new centos9 ci images (https://github.com/pulp/pulp-oci-images/pull/592), without breaking our older branches.

fixes: #843

### DIFF
--- a/templates/github/.github/workflows/scripts/install.sh.j2
+++ b/templates/github/.github/workflows/scripts/install.sh.j2
@@ -96,11 +96,11 @@ cat >> vars/main.yaml << VARSYAML
 pulp_env: {{ pulp_env | tojson }}
 pulp_settings: {{ pulp_settings | tojson }}
 pulp_scheme: {{ pulp_scheme }}
-{% if python_version == "3.6" %}
-pulp_container_tag: "python36"
+{%- if ci_base_image.count(":") %}
+pulp_default_container: {{ ci_base_image }}
 {% else %}
-pulp_container_tag: "latest"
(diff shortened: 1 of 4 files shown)


templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -96,11 +96,15 @@ cat >> vars/main.yaml << VARSYAML
...
 pulp_scheme: {{ pulp_scheme }}
-{% if python_version == "3.6" %}
-pulp_container_tag: "python36"
+{%- if ci_base_image.count(":") %}
+pulp_default_container: {{ ci_base_image }}
 {% else %}
-pulp_container_tag: "latest"
-{% endif %}
+{%- if python_version == "3.6" %}
+pulp_default_container: {{ ci_base_image + ":python36" }}
```
dralley, line 103: On what platform do we still need to care about Python 3.6?

gerrod3: No idea.

mdellweg: Let's not do this. Also let's not try to sync the python version on the outside and the inside of the container if we only do it sporadically.

mdellweg: I mean, let's drop 3.6.

mdellweg (approved):

dralley (approved):

PR #845, opened 2024-02-12
Improve python version flexibility

dralley:
* Changes the default CI image to centos9
    * Before merging let's make sure the older branches we care about have already had their template configs updated to specify the old image.  pulpcore, pulp_rpm are already done, pulp_container has PRs up, not sure about the rest
* Fixes hardcoded references to Python 3.8 in the template

[noissue]


.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "latest"
```
dralley, line 29: Note: none of these are in the template, it's the CI for the template itself

README.md
```diff
@@ -140,7 +140,7 @@ The following settings are stored in `template_config.yml`.
   pydocstyle            Boolean, whether to have flake8 use pydocstyle to check for compliance with
                         Python docstring conventions.
 
-  python_version        Python version to use in the CI. Currently only 3.6 and 3.8 are supported.
```
dralley, line 143: What restrictions on validity are there?

templates/github/.github/workflows/create-branch.yml.j2
```diff
@@ -23,7 +23,7 @@ jobs:
     steps:
       {{ checkout(depth=0, path=plugin_name) | indent(6) }}
 
-      {{ setup_python(pyversion="3.8") | indent(6) }}
```
dralley, line 26: There's no reason to use a default here.  We should use whatever the template config is set to.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -237,7 +237,7 @@ if [[ "$TEST" = "azure" ]]; then
   {%- if pulp_scheme == "https" %}
   AZCERTIFI=$(/opt/az/bin/python3 -c 'import certifi; print(certifi.where())')
   cat /usr/local/share/ca-certificates/azcert.crt >> $AZCERTIFI
-  cat /usr/local/share/ca-certificates/azcert.crt | cmd_stdin_prefix tee -a /usr/local/lib/python3.8/site-packages/certifi/cacert.pem > /dev/null
+  cat /usr/local/share/ca-certificates/azcert.crt | cmd_stdin_prefix tee -a /usr/local/lib/3.9/site-packages/certifi/cacert.pem > /dev/null
```
dralley, line 240: I believe this is specific to the *image* python version, which is a bit frustrating because it's decoupled from the `python_version` in the template config.  I'm not sure what the cleanest way to resolve that mismatch is yet.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -236,9 +236,9 @@ echo ::endgroup::
 if [[ "$TEST" = "azure" ]]; then
   {%- if pulp_scheme == "https" %}
   AZCERTIFI=$(/opt/az/bin/python3 -c 'import certifi; print(certifi.where())')
-  cat /usr/local/share/ca-certificates/azcert.crt >> $AZCERTIFI
-  cat /usr/local/share/ca-certificates/azcert.crt | cmd_stdin_prefix tee -a /usr/local/lib/python3.8/site-packages/certifi/cacert.pem > /dev/null
+  cat /usr/local/share/ca-certificates/azcert.crt | cmd_stdin_prefix tee -a $(AZCERTIFI) > /dev/null
```
dralley, line 239: AZCERTIFI seems to be the path that we wanted here.  That is, the path to the `/usr/local/lib/python3.8/site-packages/certifi/cacert.pem` where the location is correct.

templates/github/.github/workflows/scripts/install.sh.j2
dralley: see new comment

templates/bootstrap/setup.py.j2
```diff
@@ -9,11 +9,13 @@ setup(
...
     version="0.1.0a1.dev",
     description="{{ plugin_name | dash }} plugin for the Pulp Project",
+    long_description="{{ plugin_name | dash }} plugin for the Pulp Project",
+    long_description_content_type="text/markdown",
     license="GPLv2+",
     author="AUTHOR",
     author_email="author@email.here",
     url="http://example.com/",
-    python_requires=">=3.8",
+    python_requires=">=3.9",
```
dralley, line 18: I think this would be applied to old branches?  Do we need a new template variable for `minimum_supported_python`?

mdellweg: Once bootstrapped, this is a manual change. I don't think it's worth a variable.

plugin-template
```diff
@@ -67,7 +67,7 @@ DEFAULT_SETTINGS = {
     "pulp_settings": None,
     "pulp_settings_s3": None,
     "pydocstyle": True,
-    "python_version": "3.8",
+    "python_version": "3.9",
```
mdellweg, line 70: This is what we use in `setup_python` so all the CI is running on 3.9 by default?
And the python version in the container is determined by the choice of the base image, right?

.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.11"
```
mdellweg, line 29: This should be 3.9 now if i'm not mistaken.
And yes, I would stick to a python version not selecting "latest". I have been bitten too often by open ended dependencies.

README.md
```diff
@@ -140,7 +140,7 @@ The following settings are stored in `template_config.yml`.
...
+  python_version        Python version to use in the CI.
```
mdellweg, line 143:
```suggestion
  python_version        Python version to use in the CI. This does not affect the python version inside the CI container.
```

Actually, i hope, we can choose different base images in the test scenarios. -> Separate PR

mdellweg (approved):

plugin-template
dralley: Yes x2, afaik

.github/workflows/build.yml
dralley: It's just the template CI itself so it doesn't matter so much what we pick, I kinda want to stick with 3.11

mdellweg: "just". In away these are test runs for the templated workflows in the plugins. And you chose python 3.9 as a default there. Actually I would even be ok to hardcode the value of "python_version" in the `plugin-template` binary. There is really no use (anymore) for the plugin maintainer to have a choice.

It used to be the python version pulp was running on when we used the ansible based installer, but it lost that purpose a long time ago.

templates/github/.github/workflows/scripts/install.sh.j2
gerrod3: It doesn't look like you properly converted the commands. If `certifi.where() = .../certifi/cacert.pem` then the old commands do:

1. Append `azcert.crt` to `certifi/cacert.pem`
2. Write `azcert.crt` to `certific/cacert.pem` (Seems we just override step 1 if `AZCERTIFI` is equal to `python3.8/.../certifi/cacert.pem`)
3. Write `azcert.crt` to `pki/tls/cert.pem`

The new commands do:

1. Write `azcert.crt` to `certifi/cacert.pem` 
2. Write `azcert.crt` to `pki/tls/cert.pem`
3. Append `azcert.crt` to `certifi/cacert.pem` (Now this file has double copies of `azcert.crt`

The outputs of the new commands are not the same.

.github/workflows/build.yml
dralley: Would you have any objection to using 3.11 for everything other than what Pulp itself is running on?

templates/github/.github/workflows/scripts/install.sh.j2
gerrod3: My analysis was wrong, I failed to notice the `cmd_stdin_prefix` which routes the command to the container. Seeing as the pulpcore CI is still failing, the next question to ask is does `/opt/az/bin/python3` exist inside the pulp ci container? It doesn't in the oci_env, so I am suspicious that it does in the ci container.

.github/workflows/build.yml
mdellweg: No, that's ok. And I actually believe the plugin template should be in charge, because it also provides all the code that is run.

plugin-template
mdellweg: To clarify my other comment:
I would move this line to be hard coded within this file away from the DEFAULT_SETTINGS (keep it easy to parameterize, but don't let the template_config decide). You can easily inject it around line 418.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -237,7 +237,7 @@ if [[ "$TEST" = "azure" ]]; then
   {%- if pulp_scheme == "https" %}
   AZCERTIFI=$(/opt/az/bin/python3 -c 'import certifi; print(certifi.where())')
```
dralley, line 238: Gerrod and I had a call to deconstruct some of this code, we had a difficult time figuring out exactly what was going on but here's the current best idea:

* https://pypi.org/project/certifi/ is self-contained and doesn't allow the extension or modification of the trust store

* therefore this code appears to operate on the somewhat hacky principle of replacing certifi's trust store behind it's back by appending or rewriting the certificate file.  See the hacks on line 222-224

* it's doing that for both the "azcertifi" certificate on the CI host, and the certifi certificate on the CI image

.github/workflows/build.yml
dralley: Should `python_version` be fully deprecated then?

templates/github/.github/workflows/scripts/install.sh.j2
gerrod3: I tried this method for finding the correct certifi python path for the container. https://github.com/pulp/pulp_python/pull/645/files#diff-02059c243ab1ae5faa6a95ec965f0b351e0f6c96e5452c90881b6b7e3ee11618R169

.github/workflows/build.yml
mdellweg: I'd keep it in the templates and hard code in the application code so we have a single place to change it.

I think the python version comes with the image we allow to select (In an even more flexible way). I wouldn't even see a good way to assert the version pulp runs on is the one specified. And so the existence of the option raises false expectations. I may be missing your point though.

dralley: It's provided by the image we chose currently, though I know we've had various requests around making the Python version variable, and I just wasn't sure how exactly that would be implemented

But honestly that's out of scope so I'm going to drop that line of discussion

templates/github/.github/workflows/scripts/install.sh.j2
dralley: Thanks, I was trying to avoid the round trips into the container but I don't think it's going to be simpler than just doing this.

dralley:
@gerrod3 Looking at the test failure.. this line of code doesn't even exist in recent versions, we replaced aiohttp ages ago

https://github.com/pulp/plugin_template/actions/runs/7911045981/job/21594626348?pr=845#step:15:133

Any idea how this would be happening?

README.md
```diff
@@ -140,7 +140,8 @@ The following settings are stored in `template_config.yml`.
   pydocstyle            Boolean, whether to have flake8 use pydocstyle to check for compliance with
                         Python docstring conventions.
 
-  python_version        Python version to use in the CI. Currently only 3.6 and 3.8 are supported.
+  python_version        Python version to use in the CI. This does not affect the python version
+                        inside the CI container.
```
mdellweg, line 144: We can drop this line now too. Thank you for removing one more question for a useless user decision.

mdellweg (approved):
One nitpick, but also not a thing i want to hold off for.
Let's get this forward with!

gerrod3:
> @gerrod3 Looking at the test failure.. this line of code doesn't even exist in recent versions, we replaced aiohttp a full year ago
> 
> https://github.com/pulp/plugin_template/actions/runs/7911045981/job/21594626348?pr=845#step:15:133
> 
> https://github.com/pulp/pulpcore/pull/3575/files
> 
> Any idea how this would be happening?

We need to update this bootstrap line: https://github.com/pulp/plugin_template/blob/main/templates/bootstrap/requirements.txt#L1
CatDog is getting bootstrapped with a super old version of pulpcore that uses old requirements that are not compatible with python 3.11

dralley:
@gerrod3 @mdellweg If we want to keep the older branches of various plugins using the older CI images, and I assume we do if our goal is maximal stability, then we need to do a round of CI updates to set the `ci_base_image` template var on all of the old branches.

The easiest way to do that is trigger the action right now prior to merging this PR, as it will fill out the default automatically.  This is already done for pulpcore, pulp_rpm, and pulp_container but it doesn't look like ansible or gem are updated (beyond the main branch), and I don't have permissions to trigger the actions in those repos.

gerrod3:
> The easiest way to do that is trigger the action right now prior to merging this PR, as it will fill out the default automatically. This is already done for pulpcore, pulp_rpm, and pulp_container but it doesn't look like ansible or gem are updated (beyond the main branch), and I don't have permissions to trigger the actions in those repos.

pulp_gem and pulp_ansible have been updated. We still need to update pulp_ostree and pulp_deb. And do we need to do anything for pulp_file?

dralley:
pulp_file no longer has independent CI, so no.

mdellweg (approved):

mdellweg:
> pulp_file no longer has independent CI, so no.

Stable branches do...

PR #847, opened 2024-03-06
Allow repositories without plugins

mdellweg:
As programmers we start counting at 0. If a repository can house more than one plugin, no plugin should be a valid assumption too.

[noissue]


mdellweg:
Demo PR: https://github.com/pulp/pulp-certguard/pull/376

dkliban (approved):

PR #853, opened 2024-03-21
Add new_docs build test

pedro-psb:
Adds basic testing for the new docs build, which uses pulp-docs to aggregate all docs.

The script introduced in this PR runs successfully here: https://github.com/pedro-psb/pulp-docs/actions/runs/8375656897/job/22933505957#step:8:787

The other general failures are due to the fact pulp-docs doesn't have a pulp instance running and I'm not sure how to disable that with plugin_template yet.

[noissue]


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -37,6 +37,12 @@ if [[ "$TEST" = "docs" ]]; then
   exit
 fi
 
+if [[ "$TEST" = "new_docs" ]]; then
```
mdellweg, line 40: It's probably easier to just run this with the other "docs" test runner.
And in the long run we only want this one as "docs" anyway.

mdellweg (commented):
I like the idea behind this pretty much!
But i think we need some more template programming here.

We should add a flag to the template_config specifying whether staging_docs exist on this branch and only add the new test in case it was set (our old release branches would probably fail otherwise).
Maybe we want to add a second flag right away for fading out building the old docs.

pedro-psb:
The [test PR](https://github.com/pulp/pulpcore/pull/5145) is failing because pulp-docs is trying to copy protected content from `.ci/` which is not really necessary for the docs. I need to get this fixed first.

Any other repo/branch I should test on?

plugin-template
```diff
@@ -82,6 +82,8 @@ DEFAULT_SETTINGS = {
     "test_reroute": True,
     "test_s3": False,
     "use_issue_template": True,
+    "use_unified_docs": False,
+    "use_legacy_docs": True,
```
mdellweg, line 86: Please keep the alphabetical order. Also add something to README.

templates/docs/doc_requirements.txt.j2
```diff
@@ -8,6 +11,14 @@ sphinxcontrib-openapi
 towncrier
```
mdellweg, line 11: towncrier should be part of both ways of building.

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -23,6 +23,12 @@ export PULP_SETTINGS=$PWD/.ci/ansible/settings/settings.py
...
 if [[ "$TEST" = "docs" ]]; then
+{%- if use_unified_docs %}
+  # Unified Docs Build
+  pulp-docs build
+{%- endif %}
+{%- if use_legacy_docs %}
+  # Legacy Docs Build
   if [[ "$GITHUB_WORKFLOW" == "{{ plugin_app_label | camel | default("Pulp") }} CI" ]]; then
     towncrier build --yes --version 4.0.0.ci
   fi
```
mdellweg, line 34: Again, please run this in any case.

mdellweg (commented):
This looks right to me.

templates/docs/doc_requirements.txt.j2
```diff
@@ -1,13 +1,24 @@
...
 Jinja2<3.2
+{%- endif %}
+{%- if use_unified_docs %}
+
+# Unified docs
+pulp-docs @ git+https://github.com/pulp/pulp-docs@main
+{%- endif %}
 {%- for req in extra_docs_requirements %}
+
+# Extra requirements
```
mdellweg, line 22: Cosmetic change...

```suggestion
# Extra requirements
{%- for req in extra_docs_requirements %}
```

plugin-template
mdellweg: Want to document them in the README still?

mdellweg (approved):
Two nitpicks, nothing that should block this PR.

mdellweg (approved):

PR #856, opened 2024-04-03
Add ci_constraints

mdellweg:
[noisseu]


templates/github/.ci/assets/ci_constraints.txt
```diff
@@ -0,0 +1,2 @@
+tablib!=3.6.0
```
lubosmj, line 1: If `tablib` maintainers release a new version, the CI will install it and we will run into the very same issue again, or not? Does it make sense to introduce an upper bound here, like `tablib<3.6.0`?

mdellweg: It's the other way around: they released a fixed version.
So we know this one is the one version we want to avoid installing ever. And I think that is how this file should be used.

But it's unclear if we ever want to merge this.

ggainey (approved):

lubosmj (approved):

lubosmj:
If there are no objects from other folks, I am willing to merge this change...

mdellweg:
I used this to try it:
https://github.com/pulp/pulp-certguard/pull/413

PR #857, opened 2024-04-03
Update black to 24.3.0

mdellweg:
[noissue]


lubosmj (approved):
I expect Armageddon.

https://github.com/pulp/pulp_container/actions/runs/7780768444/job/21214019797

Not feeling brave to merge it immediately.

mdellweg:
We do this to avoid Armageddon: https://github.com/pulp/plugin_template/blob/main/scripts/update_ci.sh#L45

PR #859, opened 2024-04-23
Fix dependency issues triggered when installing the openapi extension

lubosmj:
This removal fixes the following error:
Could not import extension sphinxcontrib.openapi (exception: cannot import name 'error_string' from 'docutils.io' (/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/docutils/io.py))

[noissue]


lubosmj:
Tested by https://github.com/pulp/pulp_container/pull/1599. Verified locally that the REST API docs are still generated.

ggainey (approved):
What could possibly go worng? Approved! :)

mdellweg (approved):
Nice and concise.

PR #860, opened 2024-04-23
Add legacy docs deprecation warning on every page

pedro-psb:
The old docs should be deactivated on July 2024, as shown [here](https://github.com/pulp/pulp-docs/issues/33).

This should be added manually for plugin which doesn't use plugin_template docs updates.

## Preview

It looks like this (on every page):

image

[noissue]


templates/docs/docs/conf.py.j2
```diff
@@ -317,3 +317,8 @@ class MyPythonDomain(PythonDomain):
 def setup(sphinx):
     """Use MyPythonDomain in place of PythonDomain"""
     sphinx.add_domain(MyPythonDomain, override=True)
+
+rst_prolog = """.. attention::
+    This docs will be deactivated in July 2024. `Learn More <https://discourse.pulpproject.org/t/unified-documentation-part-2/1188>`_
```
dkliban, line 322: Let's just say "This documentation will be deactivated in the near future." so we don't have to do anything on a specific date.

mdellweg: Yep. Never write docs that plan to be outdated.

mdellweg (approved):

PR #861, opened 2024-04-29
Change towncrier to use markdown

pedro-psb:
This does:
- Replace .TEMPLATE.rst with .TEMPLATE.md to be used in the 'build' cmd
- Update boostrap to start using markdown instead of rst

Together with this change, I will need to manually update plugins:
- change the towncrier config on their `pyproject.toml` to use markdown
- convert the CHANGES.rst to markdown

Also, I'm not confident about how that will play with the existing backporting mechanism. Any special care I should take here?


pedro-psb:
> Do i see that correctly, that you merely copied the files over from pulp-cli?

I copied it, ran some test on fragment of pulp_rpm and did some very small adjustments, such as using `###` instead of `####` and changing the spacing between section to one space.

> Also have you checked that the changelog pickup job (.ci/scripts/collect_changes.py or so) works correctly on markdown files?

I've seem this file but I didn't test it. I'll ensure it works.

templates/github/CHANGES/.TEMPLATE.md
mdellweg, line 1: We should add the old filename to `DEPRECATED_FILES`.

https://github.com/pulp/plugin_template/blob/main/plugin-template#L90

pedro-psb:
@mdellweg I've not used [these lines](https://github.com/pulp/pulp-cli/blob/e8b4909cff8d05f992ce82927d7044197f9e50df/.ci/scripts/collect_changes.py#L48-L52) from pulp-cli because I wanted to preserve the current behavior of using the main branch changelog file path/name (and not look what each branch defines as its own changelog file).

The reason for doing so is that it looks simpler to ignore branches which doesn't have the CHANGES in markdown and only care about converting to markdown the supported branches, instead updating every single branch from every single plugin. This is mostly to avoid this considerable manual work, but we could do that if we have to. By doing so in rpm I got:

```
Looking for './CHANGES.md' at branch origin/3.24
No changelog found on this branch.
(...)
Looking for './CHANGES.md' at branch origin/3.1
No changelog found on this branch.
Looking for './CHANGES.md' at branch origin/3.0
No changelog found on this branch.
```

The tricky part is that is not so easy to test that. I've checked that a converted CHANGES.md sample (using [this script](https://gist.github.com/pedro-psb/f5ad6bd85b18ee98b76fb797ce649d9c)) looks correctly splitted/tokenized, so I assume using similar conversions for other branches would also work.

mdellweg:
It is true that we should only import md changelogs. And also they never "lived" in a different place. So that should be fine, I guess. However we will need to convert the changes files on the "supported branches" anyway.

templates/github/.ci/scripts/collect_changes.py.j2
```diff
@@ -1,15 +1,17 @@
+#!/bin/env python3
 {% include "header.j2" %}
 
 import itertools
 import os
 import re
+import tomllib
 
-import toml
```
mdellweg, line 7: There's a place somewhere where we install this now uneeded library.

mdellweg (approved):
LGTM

templates/github/.ci/scripts/collect_changes.py.j2
pedro-psb: I've removed it now

mdellweg:
I was about to hit merge. But I think we need to lay out a plan for the roll out.
Just throwing this on every supported branch over the weekend is probably producing a larger amount of grief.

pedro-psb:
Yes, agreed. My initial thought was to  submit all PRs for the supported branches till today or tomorrow. Each PR will basically:
- update CI w/ plugin-template 
- replace `CHANGES.rst` w/ `.md`
- update pyproject towncrier section (not updated by plugin-template)

pedro-psb:
I've created unit tests to the conversion script for development [here](https://github.com/pedro-psb/convert-changelogs/tree/main/tests). I can move and setup such test here if you think its worth it.

Also, I've added the new deps (pypandoc and tomlkit) on the [`update_ci.yml`](https://github.com/pulp/plugin_template/blob/main/templates/github/.github/workflows/update_ci.yml.j2) template. Is there anywhere else I should add those?

plugin-template
```diff
@@ -113,6 +113,7 @@ DEPRECATED_FILES = {
         ".travis",
         ".travis.yml",
         "flake8.cfg",
+        "CHANGES/.TEMPLATE.rst",
```
mdellweg, line 116: Maybe we should not do this after all. Sorry about that.

plugin-template
```diff
@@ -332,6 +333,7 @@ def main():
 
     if "github" in sections:
         migrate_pytest_pin(plugin_root_dir)
```
mdellweg, line 335: I guess we can safely stop attempting the pytest migration by now. WDYT?

utils.py
```diff
@@ -40,3 +44,180 @@ def to_snake(name):
...
+    package_name = pyproject_data["tool"]["towncrier"]["package"]  # type: ignore
+    TOWNCRIER_TOML_DATA = {
+        "package": package_name,
+        "filename": "CHANGES.md",
+        "directory": "CHANGES/",
+        "title_format": "## {version} ({project_date}) {{: #{version} }}",
+        "template": "CHANGES/.TEMPLATE.md",
+        "issue_format": Template(
+            "[#{issue}](https://github.com/pulp/$plugin_name/issues/{issue})"
+        ).substitute(plugin_name=package_name),
```
mdellweg, line 89: This will not always give the right result.
Can we hand the context for the jinja2 templates down here? There should be a repository variable in it.

utils.py
```diff
@@ -40,3 +44,180 @@ def to_snake(name):
...
+
+    Args:
+        changelog_file: The path to a rst changelog.
+    Return:
+        Path to new markdown changelog.
+    """
+    import pypandoc
+
+    # pre-process
+    with open(changes_rst, "r") as f:
```
mdellweg, line 113: Why are you mixing pathlib and open usage in the same file?

plugin-template
pedro-psb: I see its been around for about 4 months, so probably.

utils.py
pedro-psb: I believe we can't hand this to jinja (for example, add a `{{ repository_name }}` )because on the CI update the plugin-template won't touch this file (pyproject.toml), as far as I can tell.

utils.py
pedro-psb: No particular reason, I'll fix that.

utils.py
pedro-psb: What about using the plugin name derived from the repository path (as in line 58)? That should strictly match the repository name.

mdellweg: Whatever gives us the proper "repository_path" should be fine. But pulp repositories _can_ absolutely live outside of the pulp org.

plugin-template
pedro-psb: Why not?

mdellweg: You are right. My brain tricked me and thought about CHANGES.rst. That one we should delete explicitly after translating.

pedro-psb:
I've tested the new changes on my fork by mocking `update_ci.yml`.

The mock shows the latest commit changes after doing the update (instead of creating a PR).
Also, it uses this PR (this branch on my fork).

Mocked `update_ci.yml` using this PR: [link](https://github.com/pedro-psb/pulpcore/blob/66e3c228111950ee93cf4b2e0dd334c969a0ceae/.github/workflows/update_ci.yml#L29-L31)
plugin-template run: [link](https://github.com/pedro-psb/pulpcore/actions/runs/9022505726/job/24792184862#step:7:21)
Markdown Changelog diff: [link](https://github.com/pedro-psb/pulpcore/actions/runs/9022505726/job/24792184862#step:7:336)
Pyproject update diff: [link](https://github.com/pedro-psb/pulpcore/actions/runs/9022505726/job/24792184862#step:7:20532)

There is an error when trying to do on fork `3.53`, which *may be* related to the fact its a fork and doesn't contain the expected git objects that the original pulpcore. [error link](https://github.com/pedro-psb/pulpcore/actions/runs/9022505726/job/24792184862#step:8:70)

scripts/update_ci.sh
```diff
@@ -23,6 +23,7 @@ else
 fi
 
 pushd ../plugin_template
+pip install -r requirements.txt
 ./plugin-template --github $docs $plugin_name
 popd
```
pedro-psb, line 28: @mdellweg I've added the requirements here, because the workflow uses this script from plugin-template checkout, which will be up-to-date with what it needs.

That seems a resonable approach for handling plugin-template dependencies in general, wdyt?

CHANGES/+changelog-to-md.misc
```diff
@@ -0,0 +1 @@
+Changed towncrier config and template to use markdown.
```
mdellweg, line 1: This is nice, but also we never do releases here...
(No need to change.)

plugin-template
```diff
@@ -331,7 +332,11 @@ def main():
         print("\nAn updated plugin template config written to {path}.\n".format(path=file_path))
 
     if "github" in sections:
-        migrate_pytest_pin(plugin_root_dir)
```
mdellweg, line 334: Can you also delete the function we no longer call?
If we want to dig in the history of the project we still have git.

plugin-template
```diff
@@ -331,7 +332,11 @@ def main():
...
+        utils.migrate_changelog_to_markdown(
+            plugin_root_dir,
+            gh_org=config.get("github_org", "pulp"),
```
mdellweg, line 337:
```suggestion
            gh_org=config["github_org"],
```
The defaults must have been applied at this point. Also repeating the same (default) information is a bad idea.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+tomlkit
+pypandoc_binary
```
mdellweg, line 2: Please pin these to specific versions.
(~= is probably the right operator in this situation).

utils.py
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env python3
+from __future__ import annotations
```
mdellweg, line 2: I don't think we need this.

plugin-template
```diff
@@ -331,7 +332,11 @@ def main():
...
+            gh_repo=[n for n in plugin_root_dir.split("/") if n][-1],
```
mdellweg, line 338:
```suggestion
            plugin_name=config["plugin_name"],
```
We could also just pass "config" down...

plugin-template
```diff
@@ -9,13 +9,13 @@ import stat
 import subprocess
 import sys
 import textwrap
-import utils
-import yaml
 from pathlib import Path
 
+import yaml
 from jinja2 import Environment, FileSystemLoader
 from packaging.requirements import Requirement
```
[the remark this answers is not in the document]
pedro-psb: forget to check the CI errors, sorry

mdellweg (approved):

pedro-psb:
@mdellweg Can we merge this?

templates/bootstrap/CHANGES.md
```diff
@@ -0,0 +1,11 @@
+# {{ plugin_name }} Changelog
+
+[//]: # (You should *NOT* be adding new change log entries to this file, this)
+[//]: # (file is managed by towncrier. You *may* edit previous change logs to)
+[//]: # (fix problems like typo corrections or such.)
+[//]: # (To add a new change log entry, please see)
+[//]: # (https://docs.pulpproject.org/contributing/git.html#changelog-update)
```
mdellweg, line 7: Is that link up to date?

mdellweg (commented):
I think, this PR is ready to go.
Just one thing to confirm. The moment we merge this, we will stop having CHANGES.rst. And I think, this can break (legacy-)docs workflows in plugins. If you can confirm, we are ready to do that (and are prepared to run a pr-marathon on Monday), go ahead and merge.

templates/bootstrap/CHANGES.md
pedro-psb: No, its not

utils.py
```diff
@@ -40,3 +42,192 @@ def to_snake(name):
...
+    # manifest update
+    manifest_file = root_path / "MANIFEST.in"
+    if manifest_file.exists():
+        manifest_updated = manifest_file.read_text().replace("CHANGES.rst", "CHANGES.md")
+        manifest_file.write_text(manifest_updated)
+
+    # docs/changes.rst fix
+    legacy_changes_file = root_path / "docs" / "changes.rst"
+    if legacy_changes_file.exists():
+        legacy_changes_file.write_text(LEGACY_CHANGES_PLACEHOLDER_DATA)
```
pedro-psb, line 87: I did this here because some plugins don't have `update_docs` set to true, which could cause some trouble.

pedro-psb:
This test run on RPM uses this PR: https://github.com/pulp/pulp_rpm/actions/runs/9288246402

mdellweg (approved):

mdellweg:
Let's see how it goes. I think we have spent enough time looking for pitfalls now.

PR #863, opened 2024-05-13
Allow overwriting artifacts

mdellweg:
This way, workflows can be rerun.

[noissue]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,3 +50,4 @@ jobs:
           path: "pulp_catdog/dist/"
           if-no-files-found: "error"
           retention-days: 5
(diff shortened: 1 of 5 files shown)


lubosmj (approved):
Welcomed change!

https://github.com/actions/upload-artifact/blob/552bf3722c16e81001aea7db72d8cedf64eb5f68/README.md?plain=1#L105-L109

https://github.blog/2024-02-12-get-started-with-v4-of-github-actions-artifacts/#compatibility

> In v4, users lose the ability to upload to the same named artifact multiple times. Once an artifact is uploaded cannot be altered, and there cannot be multiple v4 artifacts with the same name, in the same workflow run.

PR #864, opened 2024-05-16
Use the merge artifacts action.

decko:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/publish.yml.j2
+++ b/templates/github/.github/workflows/publish.yml.j2
@@ -66,27 +66,45 @@ jobs:
       - name: "Upload python client packages"
         uses: "actions/upload-artifact@v4"
         with:
-          name: "python-client.tar"
+          name: "{{ plugin.app_label }}-python-client.tar"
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/publish.yml.j2
```diff
@@ -66,27 +66,45 @@ jobs:
       - name: "Upload python client packages"
         uses: "actions/upload-artifact@v4"
         with:
-          name: "python-client.tar"
+          name: "{{ plugin.app_label }}-python-client.tar"
           path: "{{ plugin_name }}/{{ plugin.app_label }}-python-client.tar"
           if-no-files-found: "error"
           overwrite: true
 
+      - name: "Merge Python client packages"
```
mdellweg, line 74: Can we use this instead?

https://github.com/actions/upload-artifact?tab=readme-ov-file#upload-using-multiple-paths-and-exclusions

decko: Sure.

PR #870, opened 2024-05-27
Stop publishing bindings

mdellweg:
[noissue]


mdellweg:
Experiment: https://github.com/pulp/pulp_gem/pull/268

decko:
🚀

pedro-psb:
Just to clarify what I meant, I want [this file](https://docs.pulpproject.org/pulpcore/api.json) to keep being published.

On a second thought, instead we could actually change how its published.
I'm not 100% sure, but I believe currently its published to the website as an static asset of the old docs build (or something like that, as I didn't find an explicit command that publishes it). So it would be nice if we could decouple the publishing of the `api.json` file from the docs build so we don't have to worry about it anymore.

From pulp-docs perspective, I would like to fetch those `api.json` from the internet.
Initially I wanted to have them in a repo, but I don't think that so important anymore.

mdellweg:
Thanks for reminding.

PR #872, opened 2024-06-03
Stop installing lsof and which

mdellweg:
This is currently broken on centos8 images.

[noissue]

They really should not be needed for the tests anyway. And if we really need them often they should be part of the CI image. If only occasionally, the individual plugin pre_before_script is the right place.


ggainey (approved):
I can only find which used in 2020-version of dynaconf, and pulp_installer. This shouldn't(?) cause problems today.

PR #873, opened 2024-06-05
Install packages from vault.centos.org on the old centos image

lubosmj:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/scripts/before_script.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_script.sh.j2
@@ -36,6 +36,13 @@ if [[ "${REDIS_DISABLED:-false}" == true ]]; then
   echo "The Redis service was disabled for $TEST"
 fi
 
+if [[ "{{ ci_base_image }}" = */pulp-ci-centos ]]; then
+  # centos-stream-8 was archived; start using vault.centos.org as a mirror when downloading packages
+  cmd_prefix bash -c "sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* \
+	  && sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-* \
+	  && yum update -y"
+fi
+
 if [[ -f $POST_BEFORE_SCRIPT ]]; then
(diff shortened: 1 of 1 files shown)


lubosmj:
This helps to resolve issues we have in the pulp-container's CI where we try installing additional packages, like flatpak, to the old centos image.

https://github.com/pulp/pulp_container/blob/d5316daf7d001a147cbb680939307a38506dfd1a/.github/workflows/scripts/post_before_script.sh#L3

lubosmj:
This is more or less the continuation of https://github.com/pulp/plugin_template/commit/2237db82e4316ded4da29f4217626502ac807196 where we camouflaged the `Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist` message.

lubosmj:
Currently being tested by https://github.com/pulp/pulp_container/pull/1653.

templates/github/.github/workflows/scripts/before_script.sh.j2
```diff
@@ -36,6 +36,14 @@ if [[ "${REDIS_DISABLED:-false}" == true ]]; then
   echo "The Redis service was disabled for $TEST"
 fi
 
+if [[ "{{ ci_base_image }}" = */pulp-ci-centos ]]; then
+  # centos-stream-8 was archived; start using vault.centos.org as a mirror when downloading packages
+  cmd_prefix bash -c "cd /etc/yum.repos.d/ \
```
decko, line 41: I think you don't need this line since `sed` is going directly to the files itself.

decko:
What's gonna happen with the images that already had this change? Throw an error? Nothing?

lubosmj:
> What's gonna happen with the images that already had this change? Throw an error? Nothing?

Which change do you have in mind? No image in the CI should have this change and it is more about hacking the way out of the problem.

decko:
> > What's gonna happen with the images that already had this change? Throw an error? Nothing?
> 
> Which change do you have in mind? No image in the CI should have this change and it is more about hacking the way out of the problem.

@mikedep333 submitted changes similar to this one to each supported branch. Like this https://github.com/pulp/pulp-oci-images/pull/638/files

lubosmj:
Ah! I have not seen it. How about we merge that instead? 🤔

PR #874, opened 2024-06-10
Allow plugins to disable template driven CI github workflow.

jctanner:
No-Issue


gerrod3:
Is this necessary? Can't you change `ci_trigger` to `workflow_dispatch`, so it only becomes triggerable manually?

jctanner:
I guess that is true.

PR #879, opened 2024-06-26
Enable plugin writers to install packages on the host OS

lubosmj:
[noissue]


lubosmj:
Tested by the second commit in https://github.com/pulp/pulp_ostree/pull/377.

mdellweg (approved):

PR #883, opened 2024-07-02
Clean up legacy docs scripts 

lubosmj:
[noissue]


lubosmj:
Tested by https://github.com/pulp/pulp_container/pull/1686/files.

lubosmj:
Hmm, we may want to restrict this to `use_legacy_docs: false` because of errors like:
```
Warning, treated as error:
/home/runner/work/pulp_container/pulp_container/pulp_container/docs/workflows/build-containerfile.rst:19:Include file '/home/runner/work/pulp_container/pulp_container/pulp_container/docs/_scripts/repo.sh' not found or reading it failed
```

lubosmj:
With this change, I will also try to remove legacy docs building and publishing from the CI as well. We no longer need that. I am going to introduce more commits along the way since the current one is isolated from the rest.

README.md
```diff
@@ -203,8 +187,6 @@ The following settings are stored in `template_config.yml`.
   pre_job_template      holds name and a path for a template to be included to run before jobs.
   post_job_template     holds name and a path for a template to be included to run after jobs.
   lint_requirements     Boolean (defaults True) to enable upper bound check on requirements.txt
-  use_unified_docs      Boolean (defaults False) to enable the unified docs CI (see [pulp-docs](https://github.com/pulp/pulp-docs).
```
mdellweg, line 206: This may actually still be helpful.

lubosmj: Can you elaborate? The only purpose of this was:
```
{%- if use_unified_docs %}
  # Unified Docs Build
  pulp-docs build
{%- endif %}
```
But, we did decide to use only new docs, did not we?

mdellweg: But not everyone may have them.
I think it's still new docs or no docs.

lubosmj: Do you think that running `pulp-docs build` will have any negative impact on plugins that have no docs, @pedro-psb?

pedro-psb: It is still able to build if a plugin does not contain any docs dir all.

plugin-template
```diff
@@ -105,21 +101,25 @@ DEPRECATED_FILES = {
...
         ".travis",
         ".travis.yml",
         "flake8.cfg",
         "CHANGES/.TEMPLATE.rst",
     ],
     "docs": [
-        "docs/_static/.this_makes_it_so_git_keeps_the__static_folder",
-        "docs/_static/api.json",
+        "docs/_static/",
+        "docs/_scripts/",
```
pedro-psb, line 122: Eventually we'll start renaming `staging_docs` to `docs` on plugins, thus removing all legacy docs. The dirs `_static` and `_scripts` have no special meaning anymore, but I can see someone keeping images on a `_static`, for example. It could be a surprise if CI tried to remove it.

I think we should not include any thing in the docs deprecation list.

lubosmj: I reverted this change. Thanks for noticing!

mdellweg (approved):

PR #884, opened 2024-07-03
Remove left-overs for building client bindings

lubosmj:
[noissue]

### DIFF
--- a/templates/github/.github/workflows/nightly.yml.j2
+++ b/templates/github/.github/workflows/nightly.yml.j2
@@ -136,8 +136,6 @@ jobs:
 
       {{ run_script(name="Install", file="install.sh") | indent(6) }}
 
-      {{ run_script(name="Install Python client", file="install_python_client.sh", withenv=False) | indent(6) }}
-
       {{ run_script(name="Before Script", file="before_script.sh", extra_env={"REDIS_DISABLED": "${{ contains('%s', matrix.env.TEST) }}" % (disabled_redis_runners | default(["TILT"]) | join(' '))}) | indent(6) }}
 
       {{ run_script(name="Performance Test", file="script.sh") | indent(6) }}
--- a/templates/github/.github/workflows/publish.yml.j2
+++ b/templates/github/.github/workflows/publish.yml.j2
@@ -25,102 +25,11 @@ jobs:
   build:
     uses: "./.github/workflows/build.yml"
 
-  build-bindings-docs:
-    needs:
-      - "build"
-    runs-on: "ubuntu-latest"
-    # Install scripts expect TEST to be set, 'docs' is most appropriate even though we don't run tests
-    env:
-      TEST: "docs"
-    steps:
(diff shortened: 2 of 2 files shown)


ggainey (dismissed):

templates/github/.github/workflows/nightly.yml.j2
```diff
@@ -136,7 +136,7 @@ jobs:
 
       {{ run_script(name="Install", file="install.sh") | indent(6) }}
 
-      {{ run_script(name="Install Python client", file="install_python_client.sh", withenv=False) | indent(6) }}
+      {{ run_script(name="Build Python bindings packages", file="build_python_client.sh", withenv=False) | indent(6) }}
```
mdellweg, line 139: I'm don't think we should still do this here.
This has moved to the build step. Either we do a download_artifact here or we let script.sh generate and install new ones.

templates/github/.github/workflows/publish.yml.j2
```diff
@@ -56,10 +56,10 @@ jobs:
 
       {{ run_script(name="Install", file="install.sh") | indent(6) }}
 
-      {{ run_script(name="Install Python client", file="install_python_client.sh", withenv=False) | indent(6) }}
+      {{ run_script(name="Build Python bindings packages", file="build_python_client.sh", withenv=False) | indent(6) }}
 
       {%- if deploy_client_to_rubygems %}
-      {{ run_script(name="Install Ruby client", file="install_ruby_client.sh", withenv=False) | indent(6) }}
+      {{ run_script(name="Build Ruby bindings packages", file="build_ruby_client.sh", withenv=False) | indent(6) }}
```
mdellweg, line 62: Same here. Actually the whole build-bindings-docs should have been done by build already. Can you completely remove that job from the workflow?

lubosmj: Sure, let me try that.

templates/github/.github/workflows/publish.yml.j2
```diff
@@ -56,12 +56,6 @@ jobs:
 
       {{ run_script(name="Install", file="install.sh") | indent(6) }}
 
-      {{ run_script(name="Install Python client", file="install_python_client.sh", withenv=False) | indent(6) }}
-
-      {%- if deploy_client_to_rubygems %}
-      {{ run_script(name="Install Ruby client", file="install_ruby_client.sh", withenv=False) | indent(6) }}
-      {%- endif %}
-
```
mdellweg, line 64: I think, you can/should remove the whole "build-bindings-docs" job from the workflow.

lubosmj: I, hopefully, removed the whole section. I need your expertise on verifying that.

mdellweg (approved):
Yes, I think, that is it.
Needs to be tested obviously by releasing a plugin.

lubosmj:
Let's give it a try! 😆

PR #885, opened 2024-07-08
Added script to update_ci that auto bumps pulpcore lowerbound

gerrod3:
@mdellweg Here is the script I was talking about, should make life easier.

### DIFF
--- a/scripts/update_core_lowerbound.py
+++ b/scripts/update_core_lowerbound.py
@@ -1,0 +1,53 @@
+import requests
+from yaml import safe_load
+from packaging.requirements import Requirement
+from packaging.version import parse
+
+
+core_template_url = "https://raw.githubusercontent.com/pulp/pulpcore/main/template_config.yml"
+
+
+def main():
+    request = requests.get(core_template_url)
+    if request.status_code != 200:
+        print("Failed to find supported branches, not checking lower bounds")
+        exit(0)
+
+    template = safe_load(request.content)
+    versions = {parse(v) for v in template.get("supported_release_branches", [])}
+    if versions:
+        changed = False
+        with open("requirements.txt") as f:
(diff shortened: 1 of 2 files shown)


scripts/update_core_lowerbound.py
```diff
@@ -0,0 +1,47 @@
...
+                        for spec in requirement.specifier:
+                            if spec.operator == ">=":
+                                min_version = parse(spec.version)
+                                if min_version not in versions:
+                                    # Lowerbound is not a supported branch, modify to next lowest
+                                    valid_lowerbounds = requirement.specifier.filter(versions)
+                                    if valid_lowerbounds:
+                                        new_min_version = min(valid_lowerbounds)
+                                        print(f"Lower bounds updated to >={new_min_version}")
+                                        lines[i] = line.replace(spec.version, str(new_min_version))
```
mdellweg, line 38: This is a bit dangerous. But I know, modifying the requirements in place is not exactly easy.

scripts/update_core_lowerbound.py
```diff
@@ -0,0 +1,47 @@
...
+                except ValueError:
+                    pass
+                else:
+                    if requirement.name == "pulpcore":
+                        for spec in requirement.specifier:
+                            if spec.operator == ">=":
+                                min_version = parse(spec.version)
+                                if min_version not in versions:
+                                    # Lowerbound is not a supported branch, modify to next lowest
+                                    valid_lowerbounds = requirement.specifier.filter(versions)
```
mdellweg, line 34: Yes!!! This was the check I was hoping to find. I think, we should error if no supported branch satisfies the requirement. At that point we can no longer support this branch either.

mdellweg (approved):

PR #886, opened 2024-07-10
Calculate lowerbounds as constraints

mdellweg:
The script was originally creating a replacement for requirements, but is now used as constraints. Postprocessing should not be necessary.

### DIFF
--- a/templates/github/.github/workflows/scripts/before_install.sh.j2
+++ b/templates/github/.github/workflows/scripts/before_install.sh.j2
@@ -62,7 +62,6 @@ fi
 
 if [[ "$TEST" = "lowerbounds" ]]; then
   python3 .ci/scripts/calc_deps_lowerbounds.py > lowerbounds_constraints.txt
-  sed -i 's/\[.*\]//g' lowerbounds_constraints.txt
 fi
 
 if [ -f $POST_BEFORE_INSTALL ]; then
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/before_install.sh.j2
```diff
@@ -62,7 +62,6 @@ fi
 
 if [[ "$TEST" = "lowerbounds" ]]; then
   python3 .ci/scripts/calc_deps_lowerbounds.py > lowerbounds_constraints.txt
-  sed -i 's/\[.*\]//g' lowerbounds_constraints.txt
```
lubosmj, line 65: Was this removal tested somewhere? I introduced this change in the past because pulpcore tried to install extras that cannot be used within the constraints.

https://github.com/pulp/plugin_template/commit/35613261247495a4a00a2780b5f893cff3956a95
https://github.com/pulp/pulpcore/blob/e53fd73824c1259cc025c62e38c11bc4ea4b01a0/setup.py#L26

mdellweg: Right. And now that the script outputs only constraints, not requirements anymore we should be fine.

See how the script only outputs lines it actually changed now.

lubosmj (approved):

PR #887, opened 2024-07-10
Add upper bounds constraints

mdellweg:
On the "pulp" test, we generate an upper bounds constraints file so we can know whether the latest version advertised by the requirement can actually be installed.


mdellweg:
Look at the special handling of pulpcore to find the latest supported version withing the range of allowed versions.

gerrod3 (approved):

PR #894, opened 2024-07-18
Fix building ruby bindings from the proper spec

mdellweg:


templates/github/.github/workflows/scripts/build_python_client.sh.j2
```diff
@@ -14,9 +14,7 @@ pushd ../pulp-openapi-generator
 {%- for plugin in plugins %}
 rm -rf "{{ plugin.name | snake }}-client"
 
-# We need to copy that over to be visible in the container.
```
mdellweg, line 17: This is no longer true. The gen-client script takes care of this now.
(And also stops to meddle with our version of the file...)

templates/github/.github/workflows/scripts/build_ruby_client.sh.j2
```diff
@@ -14,15 +14,7 @@ pushd ../pulp-openapi-generator
 {%- for plugin in plugins %}
 rm -rf "{{ plugin.name | snake }}-client"
 
-# We need to copy that over to be visible in the container.
-#cp "../{{ plugin_name }}/{{ plugin.app_label }}-api.json" .
-#./gen-client.sh "{{ plugin.app_label }}-api.json" "{{ plugin.app_label }}" ruby "{{ plugin.name | snake }}"
-
-# -------------
-# The generator still needs to have it called api.json at this time...
-cp "../{{ plugin_name }}/api.json" .
```
mdellweg, line 23: This was the wrong file to use.

ggainey (approved):

PR #898, opened 2024-07-22
Allow packages without plugins to build again

mdellweg:


mdellweg:
This is basically for pulp_file and pulp-certguard...

ggainey (approved):

PR #899, opened 2024-07-23
Spread Automatic CI Update over the day

mdellweg:


lubosmj (approved):
This should supposedly help mitigate reaching the rate limits when we run the same job on multiple repositories at once.

PR #900, opened 2024-07-24
Remove the Kanban board workflow

lubosmj:
[noissue]


ggainey (changes requested):
Do we want to check for kanban.yml as a "deprecated file"?

lubosmj:
Not a bad idea.

ggainey (approved):
Good deal - thanks!

mdellweg (approved):

Issue #901, opened 2024-07-25
Cleaning up docs-related stuff


pedro-psb:
The new process of building docs is quite different from the one we had before, and there are some things that doesnt make sense to be kept in the plugin-template machinery anymore:

## `ci_update_docs`

On a rough perspective, this option was useful before because each plugin docs was more independent.
Now the control of the docs process is centralized in the pulp-docs tool, plus we don't worry about building docs for older branches anymore. AFAIK, all the old machinery is already deactivated and I can't think of a reason to keep it.

If that sounds good, we should:
- remove `template/docs`
- add a deprecation notice to `--docs` and make it no-ops
- clean-up related code

## `docs_test`

Similarly, this made sense with independent docs, but that's unsafe in the current centralized approach.
If one repo decides to disable docs testing, all plugin CI are in the risk of failing if this one plugin merges something that will break the unified docs build.

This is to say that, if a plugin is part of the unified docs, it must test the docs.

### Alternative for not testing docs

If the plugin doesnt want to be part of the unified-docs, it can "unsubcribe" via a PR to remove itself from the participants list.
Currently, the docs-participant list lives in pulp-docs ([here](https://github.com/pulp/pulp-docs/blob/main/src/pulp_docs/data/repolist.yml)).

This handling of the subscription can be improved, but it would already work as it is.

An improvement idea I had is to keep such a list of participants here in plugin_template, so its more convenient to match with docs testing workflows: a plugin will only have docs workflow if it's in the list, and pulp-docs will only use that plugin if it's in that list. It wouldn't be a problem for pulp-docs to have to fetch that list from here, as its by nature an aggregation tool.

mdellweg:
That all sounds good to me.
And I'd say "What if a plugin not managed by us wants to have other means of docs?" is something that can be handled at their end.

PR #902, opened 2024-07-26
Move docs test workflow and rm obsolete docs processes

pedro-psb:
This does:

* Adds a docs test workflow in the ci.yml
* Removes docs test from test.yml workflow (which tests the plugin)
* Deprecate --docs feature from plugin-template
  * remove docs_test option
  * remove template/docs
  * remove template/bootstrap/docs legacy content
  * remove ci_update_docs
  * make it no-ops
* Update various info across files

Closes: #901

## Testing

This does not cover all workflows, only CI

- Run in a sample PR: https://github.com/pedro-psb/pulp_python/pull/2/checks
- Merging to non-main branch: https://github.com/pedro-psb/pulp_python/pull/3/checks
- Cache test: Not very meaningful time saving. Around 30mb of cached data
    - [no-cache run](https://github.com/pedro-psb/pulp_python/actions/runs/10117102647/job/27981407404?pr=2#step:5:19)
    - [cached run](https://github.com/pedro-psb/pulp_python/actions/runs/10117214305/job/27981735342#step:5:21)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -88,6 +92,7 @@ jobs:
       {%- endif %}
       - "lint"
       - "test"
+      - "docs"
```
pedro-psb, line 95: Will that fail if "docs" is skipped?

pedro-psb: yes

templates/github/.github/workflows/docs.yml.j2
```diff
@@ -0,0 +1,41 @@
...
+      - name: "Build docs"
+        run: |
+          pulp-docs build
+
+  no-test:
+    if: "!endsWith(github.base_ref, 'main')"
+    runs-on: "ubuntu-20.04"
+    steps:
+      - run: |
+          echo "Skip docs testing on non-main branches."
```
pedro-psb, line 41: I tried moving the skip to the ci.yml and got some trouble with the final "ready-to-ship" checks.
I could update the logic there to allow skipping docs (I can, if that sounds better), but this approach feels simpler after all.

templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -13,22 +13,14 @@ source .github/workflows/scripts/utils.sh
...
 # Gets set in .github/settings.yml, but doesn't seem to inherited by
 # this script.
 export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
 export PULP_SETTINGS=$PWD/.ci/ansible/settings/settings.py
 
 export PULP_URL="{{ pulp_scheme }}://pulp"
 
-if [[ "$TEST" = "docs" ]]; then
-  if [[ "$GITHUB_WORKFLOW" == "{{ plugin_app_label | camel | default("Pulp") }} CI" ]]; then
-    towncrier build --yes --version 4.0.0.ci
```
mdellweg, line 26: I wish we could keep this command somewhere.

pedro-psb: The `towncrier build` command?
On a first sight I just though it was not strictly related to the docs building process, but I can see we want to test if the PR contain valid assets for generating the changelogs.

I'll add a step in the new `docs.yml` called `Build changelog`, wdyt?

mdellweg: That sounds great. We should be using the generated changelog in the markdown build (or at least do a syntax check).

templates/github/.github/workflows/docs.yml.j2
```diff
@@ -28,7 +28,10 @@ jobs:
           key: {{ "${{ runner.os }}-pip-${{ hashFiles('pulp-docs-main-sha') }}" }}
           restore-keys: |
             {{ "${{ runner.os }}-pip-" }}
-      {{ install_python_deps(["-r", "doc_requirements.txt"]) | indent(6) }}
+      {{ install_python_deps(["-r", "doc_requirements.txt", "towncrier"]) | indent(6) }}
```
mdellweg, line 31: I'm inclined to ask to just make it a doc_requirement...

templates/bootstrap/doc_requirements.txt.j2
```diff
@@ -1,5 +1,2 @@
 {% include 'header.j2' %}
--r requirements.txt
-towncrier
```
mdellweg, line 3: Ohh, It was one... ;)

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -43,6 +43,9 @@ jobs:
       {%- endif %}
   {%- endif %}
 
+  docs:
+    uses: "./.github/workflows/docs.yml"
+
```
mdellweg, line 48: We _should_ keep this behind a flag, because a plugin outside of our control may want to do docs different.

pedro-psb: How do you think we should determine the docs-managed repos?
- fetching from the existing config file in pulp-docs?
- declaring a list here?
- other?

mdellweg: Oh, I see your point. So the one point of authoritative information is the pulp-docs repository, yes?
Then we should take the information from there.

templates/bootstrap/docs/index.md.j2
```diff
@@ -0,0 +1,12 @@
+# Welcome to Pulp {{ plugin_label | upper }}
+
+The `{{ plugin_label | upper }}` plugin extends pulpcore to support hosting python packages.
```
mdellweg, line 3:
```suggestion
The `{{ plugin_label | upper }}` plugin extends pulpcore to support hosting {{ plugin_label }} packages.
```

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -43,6 +43,11 @@ jobs:
       {%- endif %}
   {%- endif %}
 
+  {%- if is_pulpdocs_member %}
+  docs:
+    uses: "./.github/workflows/docs.yml"
+  {%- endif %}
+
```
mdellweg, line 50:
```suggestion
  {%- if is_pulpdocs_member %}
  
  docs:
    uses: "./.github/workflows/docs.yml"
  {%- endif %}

```

mdellweg: Or whatever it takes to get the nice extra linebreaks...
(See how the `{%-` works...)

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -16,25 +16,18 @@ PLUGIN_SOURCE="./{{ plugin_name }}/dist/{{ plugin_name | snake }}-${PLUGIN_VERSI
 export PULP_API_ROOT="{{ api_root }}"
 
 PIP_REQUIREMENTS=("{{ cli_package }}")
-if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]
-then
-  PIP_REQUIREMENTS+=("-r" "doc_requirements.txt")
 {%- if plugin_name == 'pulpcore' %}
+if [[ "$TEST" = "publish" ]]
```
mdellweg, line 20: This sounds ridiculous. There is no "publish" test.

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -191,9 +184,7 @@ fi
 
 pulp config create --base-url {{ pulp_scheme }}://pulp{% if pulp_scheme != 'https' %} --no-verify-ssl{% endif %} --api-root "$PULP_API_ROOT" --username "admin" --password "password"
 {% if test_cli -%}
-if [[ "$TEST" != "docs" ]]; then
-  cp ~/.config/pulp/cli.toml "${REPO_ROOT}/../{{ cli_package }}/tests/cli.toml"
-fi
+cp ~/.config/pulp/cli.toml "${REPO_ROOT}/../{{ cli_package }}/tests/cli.toml"
 {%- endif %}
```
mdellweg, line 188: Can you move that whole block up to around line 30 to have the cli stuff in one place?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 tomlkit~=0.12.4
 pypandoc_binary~=1.13
```
mdellweg, line 2: looks like this can be dropped again.

templates/github/.github/workflows/scripts/install.sh.j2
pedro-psb: It's safe to remove this, then?

templates/github/.github/workflows/scripts/install.sh.j2
```diff
@@ -25,7 +25,9 @@ fi
 
 pip install ${PIP_REQUIREMENTS[*]}
 
+pulp config create --base-url {{ pulp_scheme }}://pulp{% if pulp_scheme != 'https' %} --no-verify-ssl{% endif %} --api-root "$PULP_API_ROOT" --username "admin" --password "password"
 {% if test_cli -%}
+cp ~/.config/pulp/cli.toml "${REPO_ROOT}/../{{ cli_package }}/tests/cli.toml"
 PULP_CLI_VERSION="$(pip freeze | sed -n -e 's/{{ cli_package }}==//p')"
 git clone --depth 1 --branch "$PULP_CLI_VERSION" {{ cli_repo }} ../{{ cli_package }}
 {%- endif %}
```
pedro-psb, line 33: Hmm, I've just reviewed that move, around line 138 we have:

```
  {%- if test_reroute %}
  export PULP_API_ROOT="/rerouted/djnd/"
  {%- endif %}
```

So I guess it really should be down there, so it can use the proper value of PULP_API_ROOT.

mdellweg: Hmm. that explains why we do it down there.
Let's not do this now then. But we should think about a more structured approach to set all this up.

pedro-psb: Reverted

templates/github/.github/workflows/docs.yml.j2
mdellweg: I think I ran into the exact same issues before. Still anyone with a better idea...

templates/github/.github/workflows/scripts/install.sh.j2
mdellweg: I believe so. Maybe start with a "git grep publish". But I'm pretty sure this is a leftover from former versions of the release process.

templates/github/.github/workflows/scripts/install.sh.j2
mdellweg: Didn't seem to work. Don't try this again just now.

utils.py
```diff
@@ -44,190 +42,19 @@ def to_snake(name):
...
-        [//]: # (To add a new change log entry, please see the contributing docs.)
-        [//]: # (WARNING: Don't drop the towncrier directive!)
-
-        [//]: # (towncrier release notes start)
-
-        """
-    )
-    document = header + document
-    return document
+    return [line.strip()[8:] for line in response.content.decode().split("\n") if "- name:" in line]
```
mdellweg, line 60: This is a rather crude way to parse yaml...

plugin-template
```diff
@@ -328,26 +327,25 @@ def main():
     except Exception:
         config["current_version"] = "0.1.0a1.dev"
 
+    # Determine if plugin is a member of Pulp managed documentation
+    config["is_pulpdocs_member"] = config["plugin_name"] in utils.get_pulpdocs_members()
```
mdellweg, line 331: I like it. When onboarding a new plugin to the docs, we need to be extra careful. (Just a reminder at this point.)

mdellweg (approved):
This is a huge change. And maybe we can iterate to improve on it. But not necessarily in this PR.

Things to follow up on:
  - A more structured approach to define pulp settings, configure the cli, start the containers, ...
  - Are there we should mark as deprecated from the old docs templates? Is CI now the right place to delete them?

utils.py
pedro-psb: My plan is to do the minimum working solution now and later improve this membership thing (on pulp-docs, also). But I can add a yml library as requirement and make it more robust. Would you prefer that?

mdellweg: Will it stay parsing yaml? Or did you plan something completely different?

pedro-psb: I thought about doing a simple requirement-like file which lists `gh-org/gh-repo` and has `#` comments, but not sure if it's a good idea yet. If not, I might stick with yaml.

mdellweg:
Let's merge and continue from there.

PR #906, opened 2024-08-12
Add pulp_openapi_generator_ref

mdellweg:


plugin-template
```diff
@@ -52,6 +52,7 @@ DEFAULT_SETTINGS = {
     "pulp_env_azure": {},
     "pulp_env_gcp": {},
     "pulp_env_s3": {},
+    "pulp_openapi_generator_ref": None,
```
mdellweg, line 55: How do we want to go about updating this?
Shall we try to keep it floating on latest main, or find the latest available tag, and only pin it down once we branch of a release branch?
Providing a default tag here will not really help updating it in the plugins, right?

gerrod3: I like the floating on main and then pin once we branch a release branch. Sounds reasonable and straight-forward.

mdellweg: Now we have a problem: After using the default branch on "main", this would pin the release branch to a probably older commit. That does not seem quite right, no?

mdellweg:
Now I wonder if it's a good idea to let the plugin decide on the template version.
Maybe the compatibility is much more tied to the pulpcore version.

pedro-psb:
> Now I wonder if it's a good idea to let the plugin decide on the template version.
Maybe the compatibility is much more tied to the pulpcore version.

I would go further and align the openapi-generator pin with the plugin breaking changes.
I feel this was proposed already, but cant remember.

If we change this too often (and pulpcore releases too often), the chances of something going wrong between pulpcore and plugins greatly adds up. And there is the problem of a bump requiring code changes that are incompatible between generator versions, so having an unambigous agreement between core and plugins on where generator-related fixes can go looks appealing for maintainability.

mdellweg:
Yes, this is exactly the current plan. The generator will look for the pulpcore version and decide (aligned to breaking changes releases) which images to use.
I will close this PR as the road not taken.

PR #912, opened 2024-09-11
Make pytest timeout configurable and with higher defaults

pedro-psb:
As stated [here](https://github.com/pulp/plugin_template/commit/ece49c9e91e49490ba83507341b801e4ded581da), the timeout is meant to catch unexpected hanging tasks and have to meaningful stacktrace.

Then, it is not so important to minimize the timeout as it is to prevent long testing from failing unexpectedly, so I think we can raise the bar here.

01h sounds safe in general?

## Context

[This rpm test](https://github.com/pulp/pulp_rpm/actions/runs/10812774961/job/29995301298#step:15:3912) was failing due to the timeout and took me some time to realize this was this pytest-timeout plugin. On my machine it ran in 38m.

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -58,6 +58,7 @@ DEFAULT_SETTINGS = {
     "pulp_settings": None,
     "pulp_settings_s3": None,
     "pydocstyle": True,
+    "pytest_timeout": 3600,
     "release_email": "pulp-infra@redhat.com",
     "release_user": "pulpbot",
     "stalebot_days_until_close": 30,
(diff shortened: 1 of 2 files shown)


mdellweg:
You can use `@pytest.mark.timeout(3600)` on a specific test. I would prefer that.
Remember the timeout is per test, not for the whole testsuite. So just five tests hanging for 1h would hit the GH timelimit of 5h again.

pedro-psb:
Sounds fair. I did consider overriding the specific test, but was not sure it was the best option.
The GH limit is a good factor.

PR #914, opened 2024-09-16
Reorder pip install to better prioritize

mdellweg:
pipdeptree as a debugging tool should really go last.


mdellweg:
> pipdeptree as a debugging tool should really go last.

Does not seem to work.

mdellweg:
Pip’s current implementation of the provider implements get_preference as follows:

    Prefer if any of the known requirements is “direct”, e.g. points to an explicit URL.

    If equal, prefer if any requirement is “pinned”, i.e. contains operator === or ==.

    If equal, calculate an approximate “depth” and resolve requirements closer to the user-specified requirements first.

    Order user-specified requirements by the order they are specified.

    If equal, prefers “non-free” requirements, i.e. contains at least one operator, such as >= or <.

    If equal, order alphabetically for consistency (helps debuggability).

ggainey (approved):

PR #915, opened 2024-09-25
Modify reason logic for release script to show all reasons for z release

gerrod3:


templates/github/.ci/scripts/check_release.py.j2
```diff
@@ -82,12 +82,13 @@ def main():
                         branch
                     ), "Current-version has to belong to the current branch!"
                     next_version = Version(f"{branch}.{curr_version.micro + 1}")
-                    reason = "CHANGES" if z_changelog else "requirements.txt"
+                    grouped = (("Backports", z_changelog), ("requirements.txt", req_txt_diff))
+                    reasons = [r for r, c in grouped if c]
```
mdellweg, line 86: If you moved that calculation to before the if, we can have it's result as the if condition.
Would make it easier to add more reasons.

mdellweg: Or we start the loop over branches with an empty "reasons", then check things one by one and add found reasons as we go.

gerrod3: I don't understand your second point.

mdellweg: Let me try to express myself in code:
```python
reasons = []
if some_check:
    reasons.append(SOME_REASON)
if some_other_check:
    reasons.append(SOME_OTHER_REASON)
...
if reasons:
    print(f"Hey, we have reasons: {reasons}")
```

I think you would call this "The control flow goes with the data."

mdellweg: I used to adopt this pattern whenever I want to collect warnings and errors over multiple operations.

mdellweg (approved):

PR #918, opened 2024-10-07
Only use published/prebuilt bindings in the 'pulp' test scenario

### LINKED ISSUE #917: Client generated with different pulpcore breaks lowerbound tests
## The problem

Recently on rpm CI we got an error [on the lowerbound tests](https://github.com/pulp/pulp_rpm/actions/runs/11147495715/job/31044194248#step:15:6367) related to the [introduction of prns on pulpcore==3.63.0](https://pulpproject.org/pulpcore/changes/#3.63.0).

The problem is that the lowerbound uses pulpcore~=3.49, which means it shouldnt have any prn stuff in this public API at all. And that's what the bindings diff says here:

```bash
$ # build-api.json was generated in the build step based on rpm installation like  "pip install rpm -c ci_constraints.txt"
$ # test-api.json was generated in workflows/scripts/script.sh, based on container env, which has also "-c lowebound_contraints.txt"
$ jsondiff --indent 2 build-api.json test-api.json || true
{
  "op": "remove",
  "path": "/components/schemas/rpm.ModulemdObsoleteResponse/properties/prn"
},
{
  "op": "remove",
  "path": "/components/schemas/rpm.ModulemdResponse/properties/prn"
},
```

So we are using the built-client (the client that was built in the build.yml step), which was generated using a different pulpcore version than the one in our test environement installation. I've heard of this problem several times, but had never experienced such a concrete example. Yet another "that's why we are stop publishing bindings".

## Considerations

[This code ](https://github.com/pulp/pulp_rpm/blob/7a1d1a206efc6682029a003bc14626d24366c4dc/.github/workflows/scripts/script.sh#L47-L84) gives some warnings about this problem, and its where the client package is being selected.

But we can't know what pulpcore was used in the built-client (or its schema) at this point to make some kind of comparision because it doesnt show anywhere in the schema or package metadata (which is unfortunate, and maybe we should add this information on the schema, at least). 

Anyway, I'm not sure what approach to take here. I'm assuming we want those to match, so maybe we could always use the client generate there (in the `scripts/script.sh`)?

pedro-psb:
This is to prevent using clients built from different pulpcore versions than those installed in the app we want to test.

Closes #917

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -41,12 +41,14 @@ pushd ../pulp-openapi-generator
 
   # Workaround: Domains are not supported by the published bindings.
   # Sadly: Different pulpcore-versions aren't either...
-  # So we exclude the prebuilt ones only for domains disabled.
-  if [ "$(jq -r '.domain_enabled' <<<"${REPORTED_STATUS}")" = "true" ] || [ "$(jq -r '.online_workers[0].pulp_href|startswith("/pulp/api/v3/")' <<< "${REPORTED_STATUS}")" = "false" ]
+  # * In the 'pulp' scenario we use the published/prebuilt bindings, so we can test it.
+  # * In other scenarios we generate new bindings from server spec, so we have a more
+  #   reliable client.
+  if [ "$TEST" = "pulp" ]
   then
-    BUILT_CLIENTS=""
-  else
     BUILT_CLIENTS="{% for plugin in plugins %} {{ plugin.app_label }}{% endfor %} "
+  else
+    BUILT_CLIENTS=""
   fi
 
   for ITEM in $(jq -r '.versions[] | tojson' <<<"${REPORTED_STATUS}")


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -41,12 +41,14 @@ pushd ../pulp-openapi-generator
 
   # Workaround: Domains are not supported by the published bindings.
   # Sadly: Different pulpcore-versions aren't either...
-  # So we exclude the prebuilt ones only for domains disabled.
-  if [ "$(jq -r '.domain_enabled' <<<"${REPORTED_STATUS}")" = "true" ] || [ "$(jq -r '.online_workers[0].pulp_href|startswith("/pulp/api/v3/")' <<< "${REPORTED_STATUS}")" = "false" ]
+  # * In the 'pulp' scenario we use the published/prebuilt bindings, so we can test it.
+  # * In other scenarios we generate new bindings from server spec, so we have a more
+  #   reliable client.
+  if [ "$TEST" = "pulp" ]; then
```
mdellweg, line 47:
```suggestion
  if [ "$TEST" = "pulp" ]
```

mdellweg:
Please squash the commits.

mdellweg (approved):

PR #922, opened 2024-11-13
Allow to use pyproject.toml instead of setup.py

mdellweg:


ggainey (approved):
Slowly we make the build team happier :)

PR #926, opened 2024-11-18
Cleanup .bumpversion.cfg

mdellweg:
This config file is deprecated. To support older branches, we consult it as a fallback only.

### DIFF
--- a/templates/github/.ci/scripts/check_release.py.j2
+++ b/templates/github/.ci/scripts/check_release.py.j2
@@ -5,6 +5,7 @@
 import argparse
 import re
 import os
+import tomllib
 import yaml
 from tempfile import TemporaryDirectory
 from packaging.version import Version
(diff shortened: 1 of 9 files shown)


utils.py
```diff
@@ -57,4 +80,9 @@ def get_pulpdocs_members() -> list[str]:
             "This mean we can't know if we should manage the doc-related workflows."
         )
 
-    return [line.strip()[8:] for line in response.content.decode().split("\n") if "- name:" in line]
+    repolist = yaml.safe_load(response.content.decode())
+    return [
+        repo["name"]
+        for repo in itertools.chain(*repolist["repos"].values())
+        if "subpackage_of" not in repo
+    ]
```
pedro-psb, line 88: That's better :sweat_smile:
What I want to change in the format is to have the `repo_type` and `subpackages` childs of repo, so `repos` would only have repo entries.

templates/github/.ci/scripts/check_release.py.j2
```diff
@@ -17,6 +18,17 @@ Y_CHANGELOG_EXTS = [".feature", ".removal", ".deprecation"]
...
 
+def current_version(repo):
+    try:
+        pyproject_toml = tomllib.loads(repo.git.show(f"{DEFAULT_BRANCH}:pyproject.toml"))
+        current_version = pyproject_toml["project"]["version"]
+    except Exception:
+        current_version = repo.git.grep(
+            "current_version", DEFAULT_BRANCH, "--", ".bumpversion.cfg"
+        ).split("=")[-1]
+    return Version(current_version)
```
pedro-psb, line 29: This implementation looks better than the one in `utils.py` (e.g the use of split+`Version`)
Any reason they are not  the same?

pedro-psb (commented):
Why are there implementation of `current` version?

templates/github/.ci/scripts/check_release.py.j2
mdellweg: Yeah, they don't run in the same context. This one reads the version from git without checking the default branch out. And it creates a `Version` object.
The other one reads the version from an actual file of the current branch while being in the plugin_template directory and we really only want the version string.

pedro-psb (approved):

PR #929, opened 2024-11-25
Fix current_version for all branches

mdellweg:


scripts/update_core_lowerbound.py
```diff
@@ -1,52 +1,88 @@
+#!/bin/env python
+
+import re
```
gerrod3, line 3: I don't see `re` used anywhere else. Can it be removed?

mdellweg: right.

pedro-psb (approved):

PR #933, opened 2024-11-29
Allow simple alpha markers in the bumpversion config

mdellweg:


mdellweg:
We have them in pulp_npm, which at the same time we include in the "official Container images". (We shouldn't do that.)
In short this paints over a specific itch. We also introduce the alpha marker as independent, because we don't want to start an alpha -> beta -> rc -> final cycle for each new release.

pedro-psb (approved):
I see, thanks for the context.

PR #938, opened 2024-12-10
Look for WIP in commit messages

mdellweg:


pedro-psb (commented):
I was thinking about this the other day.
Can we add "FIXUP"?
And what about making it case-insensitive and constraint to matches starting with the patterns.

pedro-psb (approved):
Thanks

PR #945, opened 2025-02-17
fix code rendering in readme

evgeni:


mdellweg (approved):

ggainey (approved):
Subtle but important :)

PR #947, opened 2025-02-18
add empty unittest_requirements template

evgeni:
script.sh assumes the file is present, but nothing creates it, resulting
in failed pipelines:

    cmd_stdin_prefix bash -c "cat > /tmp/unittest_requirements.txt" < unittest_requirements.txt
    .github/workflows/scripts/script.sh: line 103: unittest_requirements.txt: No such file or directory

Fixes: 9547cdfff1b82caf51d8c407ab3560cbae8f3d25


evgeni:
You could argue the template should not manage those files, but I think it should at least pre-create empty ones so that CI doesn't fail?

mdellweg:
I think you are right. If you move these file into the bootstrap section, they will only be created, never updated.
(I agree the template should not "manage" it, just provide a default.)

evgeni:
updated, and added the deps I needed to get things running (based on pulp_container)

mdellweg (approved):

PR #948, opened 2025-02-18
update pulpcore upper bound to <3.85

evgeni:


mdellweg:
I think I know now...
we declared the lower bound "3.40" that translates to "~=3.40" in the lower bounds constraints. Instead it should be ">=3.40.0,<3.85"...

mdellweg:
And for the sake of it let's pick 3.49 as a supported y branch on Pulpcore.

I'm running experiments with that here:
https://github.com/pulp/plugin_template/pull/949

mdellweg (approved):

PR #951, opened 2025-03-24
Cleanup

mdellweg:


ggainey (approved):

ggainey:
Sometimes an idea doesn't turn out to be as good in practice as one thinks, and ends up with side effects nobody thought of :)

PR #957, opened 2025-05-05
Stop upgrading plugins to the latest core branch

mdellweg:
The upgrade strategy when the dependecy on core is newer than any supported branch, should be to just wait until eventually one of the newer branches is declared supported. If we keep upgrading the dependency aggressively, we will likely miss the window to relax on a later supported branch.

### DIFF
--- a/scripts/update_core_lowerbound.py
+++ b/scripts/update_core_lowerbound.py
@@ -4,14 +4,13 @@ import requests
 import tomlkit
 import yaml
 from packaging.requirements import Requirement
-from packaging.version import parse
+from packaging.version import Version
 
 
-core_template_url = "https://raw.githubusercontent.com/pulp/pulpcore/main/template_config.yml"
+CORE_TEMPLATE_URL = "https://raw.githubusercontent.com/pulp/pulpcore/main/template_config.yml"
 
 
-def scan_requirement(line, supported_versions):
-    updated = False
+def scan_requirement(line: str, supported_versions: set[Version]) -> str | None:
     new_requirement = None
     try:
         requirement = Requirement(line)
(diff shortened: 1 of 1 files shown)


scripts/update_core_lowerbound.py
```diff
@@ -21,67 +20,69 @@ def scan_requirement(line, supported_versions):
...
+def scan_requirements_txt(supported_versions: set[Version]) -> None:
     changed = False
     with open("requirements.txt") as fp:
         lines = fp.readlines()
         for i, line in enumerate(lines):
             requirement, sep, comment = line.partition(" #")
-            updated, new_requirement = scan_requirement(requirement, supported_versions)
-            if updated:
+            new_requirement = scan_requirement(requirement, supported_versions)
+            if new_requirement is not None:
```
ggainey, line 70: :+1:

scripts/update_core_lowerbound.py
```diff
@@ -21,67 +20,69 @@ def scan_requirement(line, supported_versions):
...
-    request = requests.get(core_template_url)
+def main() -> None:
+    request = requests.get(CORE_TEMPLATE_URL)
     if request.status_code != 200:
         print("Failed to find supported branches, not checking lower bounds")
         exit(0)
 
     core_template = yaml.safe_load(request.content)
-    supported_versions = {parse(v) for v in core_template["supported_release_branches"]}
-    latest = core_template["latest_release_branch"]
```
ggainey, line 83: and this is the key that will keep us from updating to latest-released over and over until a new supported is added. got it.

ggainey (approved):

PR #960, opened 2025-06-10
Update pulp-docs CI to use reusable workflow

pedro-psb:


plugin-template
```diff
@@ -335,14 +335,16 @@ def write_template_section(config, name, plugin_root_dir, verbose=False):
     files_templated = 0
     files_copied = 0
 
+    PULPDOCS_BRANCH = "rewrite-as-mkdocs-plugin"
```
mdellweg, line 338: So this is where we switch them all in one go, not one after another...

PR #965, opened 2025-06-17
automatically run black on files we update via templates

evgeni:
sometimes filling in a template leads to black changes (when an URL makes a line too long, for example), so lets fix that directly when generating, avoiding a second "black" commit

### DIFF
--- a/plugin-template
+++ b/plugin-template
@@ -426,6 +426,10 @@ def generate_relative_path_set(root_dir):
             template_abs_path = os.path.join(root, file_name)
             template_relative_path = os.path.relpath(template_abs_path, root_dir)
             applicable_paths.add(template_relative_path)
+    applicable_paths = list(applicable_paths)
+    if "pyproject.toml.j2" in applicable_paths:
+        applicable_paths.remove("pyproject.toml.j2")
+        applicable_paths.insert(0, "pyproject.toml.j2")
     return applicable_paths
 
 
--- a/plugin-template
+++ b/plugin-template
@@ -5,6 +5,7 @@ import os
 import pprint
 import shlex
 import shutil
+import subprocess
 import sys
 import textwrap
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


evgeni:
This is especially annoying when the resulting diff (update ci + black) is +0-0 because the only changes that happened are "regenerate a .py file, making a line non-black compliant" followed by "black".

plugin-template
```diff
@@ -402,6 +403,15 @@ def write_template_section(config, name, plugin_root_dir, verbose=False):
...
                 print(f"Copied file: {relative_path}")
+        if config["black"] and destination_relative_path.endswith((".py", ".py.j2")):
+            if destination_relative_path.endswith(".py.j2"):
+                destination = destination_relative_path[: -len(".j2")]
+            else:
+                destination = destination_relative_path
+            destination = os.path.join(plugin_root_dir, destination)
+
+            if os.path.exists(destination):
+                subprocess.run(["black", "--quiet", destination])
```
mdellweg, line 414: I like the idea. Is it evaluating settings in pyproject.toml here?

evgeni: As pwd is the template checkout, not the plugin checkout, probably not. Should I try to make it to?

evgeni: How? :)
But in the case of bootstrap, it will be fixed in a single "black" commit at some point and everything is good (again).

Updated the code to use `cwd=plugin_root`, so it *should* pickup the settings (untested)

evgeni:
Mhh, that... broke lint :D

plugin-template
evgeni: > Or just run black on everything after all the templates.

On *everything* or *everything `generate_relative_path_set` returned*? The former would possibly reformat things that were not part of the templating set.

evgeni:
I'm reasonably happy with the result.

mdellweg (approved):
I believe i convinced myself, that this is what we want. ;)

plugin-template
evgeni: Hah, but posting this, I realized, black might not (yet) be installed when `plugin-template` runs

PR #968, opened 2025-06-18
Revert "Add pycares!=4.9.0 to ci_constraints.txt"

mdellweg:
Reverts pulp/plugin_template#962

It turned out to be more than just a single bad release of pycares. Instead, we need to carefully adjust the aiodns version range on individual release branches.


mdellweg:
Why is this no including my explanation in the commit message?

mdellweg:
@pedro-psb Can you confirm this for me?

pedro-psb (approved):
That's right.

PR #975, opened 2025-08-07
Modify settings.py.j2 storages to work with older and newer branches

pedro-psb:
Summary of how the storage setting are rendered:

- pulpcore <3.70 support only legacy storage
  * test_storages_compat_layer must be unset
  * only legacy settings are rendered
- pulpcore >=3.70,<3.85 supports both legacy and new storage
  * test_storages_compat_layer must be True
  * both types of settings are rendered (for testing)
- pulpcore >=3.85 support only new storage
  * test_storages_compat_layer must be False
  * only new settings are rendered


pedro-psb:
Need to fix https://github.com/pulp/pulpcore/pull/6838 first

PR #977, opened 2025-08-18
Stop managing scenario settings

pedro-psb:
Some settings are not compatible across different scenarios x branches.

To not overcomplicate the managing logic here, from now on each branch
must define the scenario settings which is approriate for them.

---

**Context for storage**:

Guides for using storage settings*, based on pulpcore installed:

Case 1) `<3.70`
- All scenarios must render the [legacy storage settings]

Case 2) `>=3.70,<3.85`
- To tests both work, is recommended only one scenario uses the [new storage setting].
- May be only one or the other, if necessary

Case 3) `>=3.85`
- All scenarios must render the [new storage setting]

---

\*
[legacy storage settings]: based on DEFAULT_FILE_STORAGE
[new storage settings]: based on STORAGES


templates/github/.ci/ansible/settings.py.j2.copy
```diff
@@ -26,127 +26,3 @@ API_ROOT = {{ api_root | repr }}
...
-{%- macro azure_settings(legacy) -%}
-    {%- if legacy %}
-DEFAULT_FILE_STORAGE = "storages.backends.azure_storage.AzureStorage"
-AZURE_ACCOUNT_KEY = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
-AZURE_ACCOUNT_NAME = "devstoreaccount1"
-AZURE_CONTAINER =  "pulp-test"
-AZURE_LOCATION = "pulp3"
-AZURE_OVERWRITE_FILES = True
-AZURE_URL_EXPIRATION_SECS = 120
-AZURE_CONNECTION_STRING = 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://ci-azurite:10000/devstoreaccount1;'
```
mdellweg, line 75: Do you want to "move" this information into the default template values?
(I know it only helps a newly templated plugin, but it also serves as some sort of documentation.)

mdellweg (approved):

PR #980, opened 2025-09-01
Skipt test suite for doc-only changes

pedro-psb:


pedro-psb:
Results:

## With Code Changes

The CI doesn't skip tests, as expected.

<details><summary>Doesnt skip tests</summary>
<p>
image
</p>
</details> 

## No Code Changes

The CI skips the test, but github is waiting for the skipped ones.
Probably its required to actually read the "ready-to-ship"

<details><summary>Skip tests</summary>
<p>
image
</p>
</details> 

<details><summary>But waiting pending</summary>
<p>
image
</p>
</details>

pedro-psb:
Moving the logic to ready-to-ship did not resolve the no-code changes situation.
Maybe we can run the test jobs but bypass the script run?

details><summary>PR looks like waiting checks, even if its effectively done</summary>
<p>

image


</p>
</details>

pedro-psb:
**New conclusion**: 
The check hangs when skipped because they are "required checks" by branch protection rules. This rule applies to main and all release branches. There are ways of bypassing this, but if we can bypass them, why have them in the first place? We could use only the "ready-to-ship" check and put the required logic there.

**Proposal**: 
As this is an  admin per-repository setting, I'm proposing of making this skip-doc-only-changes feature opt-in.
If a repository wants to use it, it should disable protection for those checks [0], toggle the `skip_tests_on_doc_only_changes` flag in `template_config.yml` and apply plugin-template again.

[0] Protection rules are located in: `{repo} > Settings > Branches > Branch protection rules`

<details><summary>Examples ways of bypassing but which are very messy</summary>
<p>

* creating a twin `ci.yml`, one with `ignore-path: docs` and the other with the opposite. One of them can use the job names required by the checks but do nothing.
* using conditionals, because skipping is usually considered success for those required checks, but they must be skipped very explicitly (e.g, if test doesnt run because lint didnt run, it'll hang. Also that's tricky with `test.yml` matrix, because skipping the whole `test` job is not enough, and github doesn't allow skipping individual matrix runs (which is what would work in this case).

</p>
</details>

templates/github/.github/workflows/ci.yml.j2
```diff
@@ -38,15 +38,45 @@ jobs:
...
   docs:
     uses: "./.github/workflows/docs.yml"
   {%- endif %}
 
   lint:
+    needs: check-changes
+    if: needs.check-changes.outputs.run_tests == '1'
     {%- if pre_job_template %}
-    needs: {{ pre_job_template.name }}
+    needs: [check-changes, {{ pre_job_template.name }}]
```
mdellweg, line 79:
```suggestion
    needs:
      - "check-changes"
      {%- if pre_job_template %}
      - "{{ pre_job_template.name }}"
      {%- endif %}
    if: needs.check-changes.outputs.run_tests == '1'
```
The endif below needs to be removed...

PR #985, opened 2025-09-25
Cleanup the workaround for skipping docs on nondefault branches

pedro-psb:
We were skipping docs for non-default branches using a not very elegant solution. This standadizes the processes of doing job checks on ci.yml


ggainey (approved):
:+1:

PR #991, opened 2025-10-28
Update publish workflow to use PyPI Trusted Publisher

gerrod3:
Requires adding the trusted publisher to each project + client's PyPI page. Also, the GitHub "pypi" environment we are using is autocreated once the workflow is merged and ran, but you can create it before hand and assign it permissions for when it is allowed to run. ~~In ostree I created it before hand and applied the same branch protection rules we use  to only allow the environment to run on protected branches.~~

https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments

### DIFF
--- a/templates/github/.github/workflows/publish.yml.j2
+++ b/templates/github/.github/workflows/publish.yml.j2
@@ -59,10 +59,13 @@ jobs:
       id-token: "write"
 
     steps:
+      {{ checkout(depth=1, path=plugin_name) | indent(6) }}
+
       - name: "Download Python client"
         uses: "actions/download-artifact@v4"
         with:
           name: "python-client.tar"
+          path: "{{ plugin_name }}/"
 
       - name: "Untar python client packages"
         run: |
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/publish.yml.j2
```diff
@@ -30,27 +30,21 @@ jobs:
     runs-on: "ubuntu-latest"
     needs:
       - "build"
-
-    env:
-      GITHUB_TOKEN: "{{ '${{ secrets.GITHUB_TOKEN }}' }}"
+    environment:
+      name: "pypi"
+      url: "https://pypi.org/p/{{ plugin_name | dash }}"
```
mdellweg: This environment is the new GH thing for describing the "environment for a deployment" and the deployment in our case being the "upload to pypi".
You specify url here and not for the bindings. Is it needed or not?

mdellweg: Sad, but I'll leave it to your discretion.

PR #1028, opened 2026-04-20
Switch formatting and linting to ruff

mdellweg:


mdellweg:
I think we can add import sorting in a separate go...

dralley (approved):

PR #1029, opened 2026-04-22
Add import sorting

mdellweg:
Also activates ruff lint rulesets for preventing pulpcore.app imports and stray debugger statements.


templates/github/.github/workflows/lint.yml.j2
```diff
@@ -35,12 +35,6 @@ jobs:
       - name: "Lint code"
         run: |
           ruff check
-      {%- if check_gettext %}
-
-      - name: "Check for common gettext problems"
-        run: |
-          sh .ci/scripts/check_gettext.sh
```
mdellweg, line 42: There was a set of flake8 rules for this all along... We activate them in ruff with the "INT" ruleset.

.ci/scripts/pr_labels.py
```diff
@@ -4,9 +4,9 @@
 
 import re
 import sys
-import tomllib
 from pathlib import Path
 
+import tomllib
```
dralley, line 9: This is in the stdlib so it's a bit weird that it gets sorted into the 3rd party section

dralley (approved):
No objections other than the weirdness with tomllib.  But not blocking

.ci/scripts/pr_labels.py
mdellweg: I think this tool is sensitive on the python version and tomllib was only recently added to std...

dralley: It was added in 3.11 so not *that* recent :)

PR #1030, opened 2026-04-22
Fix linting

mdellweg:
`ruff check --diff` implies `--fix-only` and therefor ignores some rules.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,7 @@ jobs:
         run: |
           pip3 install ruff
           ruff format --check --diff plugin-template utils.py
-          ruff check --diff plugin-template utils.py
+          ruff check plugin-template utils.py
       - name: "Bootstrap catdog plugin"
         working-directory: "plugin_template"
         run: |
@@ -54,7 +54,7 @@ jobs:
 
       - name: "Lint code"
         run: |
-          ruff check --diff
+          ruff check
 
       - name: "Check for common gettext problems"
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: |
           pip3 install ruff
           ruff format --check --diff plugin-template utils.py
-          ruff check --diff plugin-template utils.py
+          ruff check plugin-template utils.py
```
jobselko, line 35: Not sure if I understand the change based on the PR description. `--diff` shows the patch (diff) of what would be changed instead of applying fixes

dralley (approved):

.github/workflows/lint.yml
dralley: Honestly the documentation / descriptions are not terribly precise.  I presume that `ruff check` already does not apply fixes otherwise what would be the point of `--fix`

And for linting I guess we do want to be warned of everything (even as ruff is updated) whereas for formatting we maybe don't care.

I think `--diff` is described in relation to `--fix`, which it implies yet differs from, rather than in relation to not applying it at all.

jobselko: `ruff check` reports issues and `ruff check --fix` applies fixes. Never mind, the PR does what we need :)

dralley: Yes, I consider this a documentation issue on the part of `ruff`.  Very confusing at a glance.

mdellweg: The important part in the docs is the "--diff Implies `--fix-only`" (see `ruff check --help`).
I thought showing the diff for fixable offenses was nice, but not at the expense of ignoring unfixable ones.
Maybe "--diff --no-fix-only" would work, but at this point is see too much danger and it's even more confusing.

"--diff" for format is different, because as long as the code is syntactically correct, there is no such thing as _unformatible_ style changes.

PR #1032, opened 2026-04-26
Fix pulpcore plugin comparison

dralley:
The correct comparison is "pulpcore", not "core".


templates/github/pyproject.toml.tool.j2
```diff
@@ -74,7 +74,7 @@ extend-select = [
   "TID",
   "T10",
 ]
-{%- if plugin_name != "core" %}
+{%- if plugin_name != "pulpcore" %}
```
mdellweg, line 77: :+1:

PR #1037, opened 2026-06-08
Filter pulp-cli livetest to the plugin under test

gerrod3:
## Summary
- Pass `PYTEST_MARK="live and (<plugin markers>)"` when invoking `make livetest` for pulp-cli 0.40.0+, which no longer ships `test_requirements.txt`
- Keeps the existing path for older pulp-cli versions that install test deps from `test_requirements.txt` and run filtered pytest directly

Depends on https://github.com/pulp/pulp-cli/pull/1419, which adds the `PYTEST_MARK` variable to the pulp-cli Makefile.

## Test plan
- [ ] Regenerate CI for a plugin (e.g. pulp_container) and confirm `script.sh` uses the filtered livetest invocation
- [ ] With pulp-cli < 0.40.0 checked out, CI still uses `test_requirements.txt` + `pytest -m "<plugin>"`
- [ ] With pulp-cli 0.40.0+ checked out, CI runs `make livetest PYTEST_MARK="live and pulp_container"` (or equivalent for multi-plugin repos)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -159,13 +159,14 @@ export PULP_FIXTURES_URL="http://pulp-fixtures:8080"
 {%- endif %}
 # some pulp-cli tests use the api root envvar
 export PULP_API_ROOT="$(EDITOR=cat pulp config edit 2>/dev/null | awk -F'"' '/api_root/{print $2; exit}')"
+{%- set cli_plugin_markers = plugins | map(attribute='name') | join(' or ') | snake %}
 pushd ../{{ cli_package }}
 if [[ -f "test_requirements.txt" ]]
 then
   pip install -r test_requirements.txt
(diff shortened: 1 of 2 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -159,13 +159,14 @@ export PULP_FIXTURES_URL="http://pulp-fixtures:8080"
 {%- endif %}
 # some pulp-cli tests use the api root envvar
 export PULP_API_ROOT="$(EDITOR=cat pulp config edit 2>/dev/null | awk -F'"' '/api_root/{print $2; exit}')"
+{%- set cli_plugin_markers = plugins | map(attribute='name') | join(' or ') | snake %}
```
ggainey, line 162: 👍

ggainey (approved):

PR #1041, opened 2026-07-23
Set ruff select setting for 0.16.0

daviddavis:
The new release of ruff 0.16.0 is raising hundreds of new errors for plugins. This sets select back to what it was < 0.16.0.

Blog post with more info:

https://astral.sh/blog/ruff-v0.16.0#better-default-rule-set


mdellweg (approved):
Thank you!

I think we may want to look at some of the rules but in a different pace.

PR #1045, opened 2026-08-12
Make cli tests run in parallel

gerrod3:
https://redhat.atlassian.net/browse/PULP-2236

### 📜 Checklist

### DIFF
--- a/templates/github/.github/workflows/scripts/script.sh.j2
+++ b/templates/github/.github/workflows/scripts/script.sh.j2
@@ -161,13 +161,7 @@ export PULP_FIXTURES_URL="http://pulp-fixtures:8080"
 export PULP_API_ROOT="$(EDITOR=cat pulp config edit 2>/dev/null | awk -F'"' '/api_root/{print $2; exit}')"
 {%- set cli_plugin_markers = plugins | map(attribute='name') | join(' or ') | snake %}
 pushd ../{{ cli_package }}
-if [[ -f "test_requirements.txt" ]]
-then
-  uv pip install -r test_requirements.txt
-  pytest -v tests -m "{{ cli_plugin_markers }}"
-else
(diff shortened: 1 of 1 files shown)


templates/github/.github/workflows/scripts/script.sh.j2
```diff
@@ -166,7 +166,7 @@ then
   pip install -r test_requirements.txt
   pytest -v tests -m "{{ cli_plugin_markers }}"
 else
-  PULP_CA_BUNDLE="/usr/local/share/ca-certificates/pulp_webserver.crt" make livetest PYTEST_MARK="live and ({{ cli_plugin_markers }})"
+  PULP_CA_BUNDLE="/usr/local/share/ca-certificates/pulp_webserver.crt" make paralleltest PYTEST_MARK="live and ({{ cli_plugin_markers }})"
 fi
```
mdellweg: This looks like a "is this cli new enough" heuristic. But I believe it's not accurate anymore.
What we can do however (by the new power of uv), is ignore the version of pulp_glue installed for pulpcore completely and run the cli (always the latest release) in an isolated environment.
WDYT?

https://docs.astral.sh/uv/reference/cli/#uv-run--isolated

gerrod3: Which line are you talking about? 166? I'm not sure what you are asking for.

mdellweg: The whole if "test_requirements.txt".
Which is i think rooted in the fact that since pulpcore chooses one version of glue the one of the cli was pinned. But with an isolated environment, we can make that obstacle go away completely. And we would never need to touch an old cli branch because some plugin CI happened to install from it again.

gerrod3: The original if check was because plugins with their own CLI didn't all have the new uv logic in their repo. Maven and ostree still both don't have it. I'll go update them and then update this.

mdellweg (approved):

PR #1050, opened 2026-08-27
Derive the CI Python version from the package's requires-python.

gerrod3:
This will allow us to upgrade to require>=3.12 without worry of older branches.

### 📜 Checklist

### DIFF
--- a/templates/github/.github/workflows/ci.yml.j2
+++ b/templates/github/.github/workflows/ci.yml.j2
@@ -49,7 +49,7 @@ jobs:
     steps:
       {{ checkout(depth=0, path=plugin_name) | indent(6) }}
 
-      {{ setup_python("3.12") | indent(6) }}
+      {{ setup_python() | indent(6) }}
 
       {{ install_python_deps(["gitpython"]) | indent(6) }}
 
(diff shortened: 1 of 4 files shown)


templates/github/.github/workflows/ci.yml.j2
```diff
@@ -49,7 +49,7 @@ jobs:
     steps:
       {{ checkout(depth=0, path=plugin_name) | indent(6) }}
 
-      {{ setup_python("3.12") | indent(6) }}
+      {{ setup_python() | indent(6) }}
```
mdellweg, line 52: I think this workflow does not execute any code from the plugin. So it should be fine (and maybe wise?) to let the template decide.

mdellweg: Not a blocking requirement however...

We could also start transitioning this job to uv.

gerrod3: It already uses uv: https://github.com/pulp/plugin_template/blob/main/templates/include/macros.j2#L40

All the ci jobs should use uv now.

mdellweg: Sorry I was not wording this right. The templates use `uv pip install` as a drop in replacement (makes me wonder if you could even do `alias pip "uv pip"` these days.).
What I meant was using `uv run --script` instead, where the script tells uv it needs gitpython.

Anyway none of this ought to block this PR.

mdellweg (approved):

PR #1052, opened 2026-09-16
Add migration to the config for rustfs

mdellweg:
### 📜 Checklist


plugin-template
```diff
@@ -334,6 +335,13 @@ def load_config(plugin_config_path: Path, verbose: bool) -> tuple[dict[str, t.An
     if "flake8_ignore" in config:
         config["lint_ignore"] = config.pop("flake8_ignore")
         write_new_config = True
+    # Migrate minio to rustfs
+    with suppress(KeyError):
+        if "s3boto3" in config["pulp_settings_s3"]["STORAGES"]["default"]["BACKEND"]:
+            config["pulp_settings_s3"]["STORAGES"]["default"]["OPTIONS"]["endpoint_url"] = (
```
gerrod3, line 341: Shouldn't this also check to see if the url is minio? This migration will always run even after the switch to rustfs?

mdellweg: What even is the benefit for plugins to overwrite this config in the first place?
The parameters with which the container is started are in control of the plugin.

Otoh whatever we template here still goes through test and review on every single plugin branch. So i opt for not borrowing problems of another universe here.

(The whole scenario framework should be less centered around three select storage apis. But this is where we are today.)

gerrod3 (approved):

PR #1055, opened 2026-09-18
Use python 3.12 for ci controlled scripts

mdellweg:
### 📜 Checklist


templates/github/.github/workflows/test.yml.j2
```diff
@@ -70,7 +70,7 @@ jobs:
           {%- endfor %}
       {%- endif %}
 
-      {{ install_python_deps(["build", "towncrier", "twine", "wheel", "httpie", "docker", "netaddr", "boto3", "ansible~=10.3.0", "mkdocs", "jq", "jsonpatch", "bump-my-version"]) | indent(6) }}
+      {{ install_python_deps(["httpie", "docker", "netaddr", "boto3", "ansible~=10.3.0", "mkdocs", "jsonpatch"]) | indent(6) }}
```
gerrod3, line 73: Is this now the last invocation of `install_python_deps`?

.github/workflows/test.yml
```diff
@@ -150,7 +149,7 @@ jobs:
         if: always()
         run: |
           echo "Need to debug? Please check: https://github.com/marketplace/actions/debugging-with-tmate"
-          http --timeout 30 --check-status --pretty format --print hb "https://pulp${PULP_API_ROOT}api/v3/status/" || true
+          pulp status || true
```
gerrod3, line 152: I don't think it's guaranteed that the cli will be installed. Also, why is this changing now? I don't see the change for display_logs macro?

gerrod3 (commented):
I think we should open a test PR against pulpcore to check that this works first. It looks fine, but I wonder if installing dependencies per script run will expose that we have been relying on global dependencies installed from previous steps.

mdellweg:
> I think we should open a test PR against pulpcore to check that this works first. It looks fine, but I wonder if installing dependencies per script run will expose that we have been relying on global dependencies installed from previous steps.

I did try it on this PR:
https://github.com/pulp/pulp_ansible/pull/2632

.github/workflows/test.yml
mdellweg: Well, it is for a long time now. And this workflow file was just outdated compared to the templates.

(Apparently not everybody adjusting the templates brings the changes to these bootstrap files too.)

templates/github/.github/workflows/test.yml.j2
mdellweg: I don't know. I didn't make the uv change complete in all places. Kind of reducing the risk here.

mdellweg:
Let's add https://github.com/pulp/pulpcore/pull/8109 for the test.

gerrod3 (approved):
Pulpcore passes, so looks good to me. Thanks!

